PR #68, opened 2015-02-12
try to fix security problem

javisantana:
a query like:

```
select cdb_querytables("select\n* from pg_class")
```

returns an empty set. That means that SQL API can't check if an user is selecting data from pg_class or other system tables (which is forbidden)

Removing that regexp check and letting postgres to do its work in the EXPLAIN makes it work.

I'm wondering if we could remove permissions to system tables to publicuser.

The problem with this PR is that changes the way cdb_querytables work. Right now if you do:

```
select cdb_querytables('create table (a int)')
```
it returns [] because the "ilike select %" hack removes the create table statement. 

Questions:

- is there a way to know if a query will write?
- any solution to this?

cc @pramsey @rochoa @Kartones @zenitraM

### DIFF
--- a/scripts-available/CDB_QueryTables.sql
+++ b/scripts-available/CDB_QueryTables.sql
@@ -16,6 +16,11 @@ BEGIN
 
   FOR rec IN SELECT CDB_QueryStatements(query) q LOOP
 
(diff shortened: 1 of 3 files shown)


javisantana:
cc @rochoa

PR #70, opened 2015-02-18
Remove grant select from cdb tablemetadata to public

rochoa:
- Select permission in CDB_TableMetadata no longer granted to public
 - New function to upsert the updated_at in CDB_TableMetadata for a regclass


rochoa:
cc @Kartones @javisantana review this one

Kartones:
+1 (and sorry for the late CR)

scripts-available/CDB_TableMetadata.sql
```diff
@@ -118,3 +118,21 @@ CREATE TRIGGER table_modified AFTER INSERT OR UPDATE
 ON CDB_TableMetadata FOR EACH ROW EXECUTE PROCEDURE
  _CDB_TableMetadata_Updated();
 
+
+-- similar to TOUCH(1) in unix filesystems but for table in cdb_tablemetadata
+CREATE OR REPLACE FUNCTION public.CDB_TableMetadataTouch(tablename regclass)
```
javisantana: I understand this function will only work for roles with permissions on cdb_tablemetadata, right?

javisantana:
:es: for me, I left a question for clarification

scripts-available/CDB_TableMetadata.sql
rochoa: Correct! Anyway I'm gonna add a test to cover that behaviour for future/potential regressions.

PR #75, opened 2015-03-03
73 fix upgrade of cdb stringtodate

rafatower:
@Kartones review please


Kartones:
+1 (ouch for not spotting it me neither)

rafatower:
I'm merging now cause loose ends can bite us. Don't we have tests to check upgrades within the extension code?

PR #78, opened 2015-04-20
[wip] New cartodbfy function (overwrites CDB_CartodbfyTable)

javisantana:
Still needs to be fully tested (partially tested now) using
the existing regression tests. Does not manage the timestamp
columns at this time.


scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -649,3 +649,649 @@ BEGIN
...
+  
+    IF NOT FOUND THEN
+      RETURN newrelname;
+    END IF;
+    
+    i := i + 1;
+    newrelname := relationname || '_' || i;
+  
+    IF i > 100 THEN
+      RAISE EXCEPTION '_CDB_Unique_Relation_Name looping too far';
```
rochoa: I guess we would have to add all the new exceptions to the calling code (importer) so they can be managed, wouldn't we?

scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -649,3 +649,649 @@ BEGIN
...
+    EXECUTE sql;
+  END LOOP;
+    
+  RETURN true;
+
+END;
+$$ LANGUAGE 'plpgsql';
+
+
+CREATE OR REPLACE FUNCTION CDB_CartodbfyTable2(reloid REGCLASS, destschema TEXT DEFAULT NULL)
```
rochoa: Thinking about cartodb-postgresql being another CartoDB API, should we consider CDB_* functions the public API and any other function private / unsupported functions that you can use at your own peril?

scripts-available/CDB_CartodbfyTable.sql
pramsey: What if they all had a common format? Then the calling code could just extract the relevant message text and pass it on... right now all the exceptions are of the "I'm dead, call an ambulance" variety.

rochoa: LOL. I guess that would be pretty awesome, cc @Kartones

Kartones: We already parse some outputs from ogr2ogr/sql/etc, so perfect if we format them to distinguish this kind of "naming failures" (and similar cartodbfication issues) from fatal errors

rafatower:
FYI @pramsey I'll be working on integrating this with the editor https://github.com/CartoDB/cartodb/issues/4962

scripts-available/CDB_CartodbfyTable.sql
pramsey: We shouldn't just think it, we should *accept* it, because whatever we think, that's what's going to happen. 3rd party importers call CDB_Cartodbfy already, lots of other CDB_* functions are already public, in that they are documented and promoted by the community team. I think marking the non-public ones, with an underbar or something, makes a lot of sense.

javisantana:
one question that is open:

should we add a index to the_geom?

generating index for the_geom takes a good amount of time and I'm not so sure if it's useful at all

what is your opinion @pramsey @Kartones @rafatower @rochoa ?

pramsey:
Gotta do it. It's not useful to cartodb mainline functions, but SQL API users swap back and forth between `the_geom` based queries and `the_geom_webmercator` queries all the time. Having a surprise "your query will suck unless you add an index" case would reduce the Magic of CartoDB.

scripts-available/CDB_CartodbfyTable.sql
rochoa: @javisantana what's your take about *official/supported* functions?

I'm OK with the "whatever starts with `cdb_` is supported" approach but I guess we should start listing / documenting those functions somewhere because they are part of our platform.

javisantana: We know that CDB_Cartodbfytable and CDB_usertables are already public so let's start with those.

https://github.com/CartoDB/cartodb-postgresql/issues/105

scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -649,3 +649,837 @@ BEGIN
...
+
+  i := 0;
+  newrelname := relationname;
+  LOOP
+
+    SELECT c.relname, n.nspname
+    INTO rec
+    FROM pg_class c
+    JOIN pg_namespace n ON c.relnamespace = n.oid
+    WHERE c.relname = newrelname
```
rochoa: Is it a common practice to loop until valid name is found? Put it this way: can we filter by `c.relname ilike newrelname || '%' and decide the name based on existing names?

pramsey: Two tricks: finding only names that match the established "pattern" for these things (foo_2) and finding the "next" name. Probably a query that `regexp_matches` the current names can do both at once and spit out an answer suitable for appending the "next" number to. This isn't a particularly deep loop, but replacing loops with SQL is always a good idea, IMO.

scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -649,3 +649,837 @@ BEGIN
...
+--     As before, this drops all the metadata and geom sync triggers
+--
+-- (2) _CDB_Has_Usable_Primary_ID()
+--     Returns TRUE if it can find a unique integer primary key named
+--    'cartodb_id' or can rename an existing key.
+--     Returns FALSE otherwise.
+--
+-- (3) _CDB_Has_Usable_Geom()
+--     Looks for existing EPSG:4326 and EPSG:3857 geometry columns, and
+--     renames them to the standard names if it can find them, returning TRUE.
```
rochoa: Is this the desired behaviour? Or do we want to keep the original columns?

I know it's a non-sense to have columns duplicated but as I use I can find myself lost because some of my columns disappeared.

pramsey: This is what happens now, AFAIK.

scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -649,3 +649,837 @@ BEGIN
...
+-- (4) _CDB_Rewrite_Table()
+--     If table does not have a usable primary key and both usable geom
+--     columns it needs to be re-written. Function constructs an appropriate
+--     CREATE TABLE AS SELECT... query and executes it.
+--
+-- (5) _CDB_Add_Indexes()
+--     Checks the primary key column for primary key constraint, adds it if
+--     missing. Check geometry columns for GIST indexes and adds them if missing.
+--
+-- (6) _CDB_create_triggers()
```
rochoa: I guess it's not implemented yet, is it?

scripts-available/CDB_CartodbfyTable.sql
rochoa: fair enough :laughing:

scripts-available/CDB_CartodbfyTable.sql
pramsey: Implemented (in old implementation, as is `_CDB_drop_triggers()`) but not enabled... can't remember why, possibly for testing purposes during dev. Will have to check that everything works right with it turned on (almost certainly).

rafatower: the call is commented out, but the legacy `_CDB_create_triggers()` is implemented in the same file above. I'll check whether it needs tweaking or not.

scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -649,3 +649,837 @@ BEGIN
...
+  -- the sequence does too
+  sql := Format('ALTER SEQUENCE %s OWNED BY %s.%s', destseq, copyname, const.pkey);
+  PERFORM _CDB_SQL(sql,'_CDB_Rewrite_Table');
+  
+  -- We just made a copy, so we can drop the original now
+  sql := Format('DROP TABLE %s', reloid::text);
+  PERFORM _CDB_SQL(sql, '_CDB_Rewrite_Table');
+  
+  -- If we used a temporary destination table
+  -- we can now rename it into place
```
rafatower: what happens otherwise? is it the caller's responsibility?

pramsey: The re-write will have created the new table right into the destination location we wanted.  The case of the no-write-required will have already returned higher up in the function, so at this point we know we've re-written the table. Question is, did we rewrite it to a temp location or a permanent location? if the latter, then we've nothing left to do. By the time this function returns, we guarantee there's a good table in the right place, so no responsibility handed to caller.

scripts-available/CDB_CartodbfyTable.sql
pramsey: Confirmed, just enabling it works fine.

pramsey:
Important note: the current code handles raster tables, this code doesn't.

rafatower:
We also need to account for raster tables

pramsey:
Mostly the current raster support just seems to be adding triggers and assuming that all columns are already correctly named, so it's pretty brittle.

rafatower:
Just got this error testing everything together with the triggers enabled.
```
2015-08-10 16:10:34 UTC LOG:  statement: 
                SELECT cartodb.CDB_CartodbfyTable2('five_countries_15', 'public');
              
2015-08-10 16:10:34 UTC ERROR:  syntax error at or near "8022947" at character 78
2015-08-10 16:10:34 UTC QUERY:  CREATE trigger track_updates AFTER INSERT OR UPDATE OR DELETE OR TRUNCATE ON 8022947 FOR EACH STATEMENT EXECUTE PROCEDURE cartodb.cdb_tablemetadata_trigger()
2015-08-10 16:10:34 UTC CONTEXT:  PL/pgSQL function _cdb_create_triggers(text,regclass) line 9 at EXECUTE statement
        SQL statement "SELECT _CDB_create_triggers(destschema, reloid)"
        PL/pgSQL function cdb_cartodbfytable2(regclass,text) line 47 at PERFORM
2015-08-10 16:10:34 UTC STATEMENT:  
                SELECT cartodb.CDB_CartodbfyTable2('five_countries_15', 'public');
```

pramsey:
Sorry, it's not just enabling the call, it's changing one param too, so it looks like
```
  -- Add triggers to the destination table, as necessary
  PERFORM _CDB_create_triggers(destschema, destoid);
```

rafatower:
thanks for pointing out so quickly :)

pramsey:
For raster tables, just copying the logic from the existing code looks like this
```
  -- Rasters only get a cartodb_id and a limited selection of triggers
  -- underlying assumption is that they are already formed up correctly
  SELECT cartodb._CDB_is_raster_table(schema_name, reloid) INTO is_raster;
  IF is_raster THEN

    PERFORM cartodb._CDB_create_cartodb_id_column(reloid);
    PERFORM cartodb._CDB_create_raster_triggers(destschema, reloid);

  ELSE
    
    -- Rewrite (or rename) the table to the new location
    PERFORM _CDB_Rewrite_Table(reloid, destschema);

    -- The old regclass might not be valid anymore if we re-wrote the table...
    destoid := (destschema || '.' || destname)::regclass;

    -- Add indexes to the destination table, as necessary
    PERFORM _CDB_Add_Indexes(destoid);
  
    -- Add triggers to the destination table, as necessary
    PERFORM _CDB_create_triggers(destschema, destoid);

  END IF;
```
I'm not sure how happy we really are w/ this, probably we should do the full song-and-dance with rasters, in terms of ensuring that columns have the right SRS, right raster column names, etc, etc, etc. There's a whole extra layer for raster though, which is there's a hard expectation of certain overview raster tables existing, no?

rafatower:
> I'm not sure how happy we really are w/ this, probably we should do the full song-and-dance with rasters...

I guess it depends on roadmap. I think we can go ahead with the copied logic as you suggest, but summoning @javisantana just in case.

scripts-available/CDB_CartodbfyTable.sql
pramsey: I tried a `regexp_match` approach, and it's really ugly and starts with an identity ("does this relation already exist?") query anyways, so I think the simplicity of the loops is better to stick with.
```
WITH raw AS (SELECT c.relname, n.nspname, ( select regexp_matches(c.relname, E'bar_(\\d+)') )[1]::integer + 1 as c
    FROM pg_class c
    JOIN pg_namespace n ON c.relnamespace = n.oid
    WHERE ( c.relname = 'bar' OR c.relname ~ ('bar' || E'_\\d+$') )
    AND n.nspname = 'public' order by c desc) SELECT relname, nspname, case when c is null then 0 else c end as c from raw order by c desc;
```

rochoa: I'm OK, just wanted to know how ugly it could be the other approach. Now we know it so let's go with the loop :+1:

scripts-available/CDB_CartodbfyTable.sql
Kartones: Yep at least in theory this is what cartodbfy does now... if everything looks fine it keeps geom columns, else renames them and creates proper ones from scratch

javisantana:
@pramsey could you add the raster code to the PR? 
could you also check that new function works with ogr2ogr? (I'm thinking here on FME)

rafatower:
I have to check a number of things before deciding on the best deploy approach. I just found this:
```
exception:  PG::Error: ERROR:  record "new" has no field "updated_at"
CONTEXT:  PL/pgSQL function _cdb_update_updated_at() line 3 at assignment

PR #80, opened 2015-04-28
updates quantile bins algorithm

ohasselblad:
Doing a code review of #79 I noticed a few more problems with this algorithm:

1. Removes `group by` clause that shouldn't be included (all values should be included, not only unique ones)
2. Adds cases to handle non integer bin positions (halfway between adjacent points)

cc @javisantana


sariogonfer:
1. If 'group by' clause is removed, this can provoke error in scenarios where one value repeats multiple times, and this value would appear as limit of several bins.
2. I think the last value of the bin is more useful than a halfway between adjacent points, in case you want to show this value inside a legend, for example

ohasselblad:
For 1., this is common behavior for quantile bins functions. For instance, R produces the following when calculating the [quantile breaks](https://stat.ethz.ch/R-manual/R-patched/library/stats/html/quantile.html):
screen shot 2015-04-28 at 15 26 49

I get the same values returned by the function here.

For 2., I agree that it is arbitrary choosing the halfway point, but [it is conventional](http://en.wikipedia.org/wiki/Quantile#Quantiles_of_a_population).

ohasselblad:
We could add an option of choosing the end point instead of the halfway point. What do you think @sariogonfer ?

sariogonfer:
That is a good idea, i think.
Other possibility is to do two function, one of them will return the strict definition of quantile bin, and the other one return a result more useful if you want to work with maps (with group by clause and return the biggest value of the quantile).

ohasselblad:
bump @javisantana

ohasselblad:
@sariogonfer we're going to go with this version for now. it effectively reduces the cartocss for the choropleth to what one would see with doing unique values, and the legend only shows the extreme values so that won't be effected by doing the `avg`. We're going to leave it to the user to customize the CartoCSS further if they are not happy with the values returned by this function.

Thanks so much for finding the errors before!

PR #83, opened 2015-06-01
Replace CDB_UserTables with something that can handle multi-user accounts

pramsey:


pramsey:
This is ready to be merged, IMO, @rochoa and maybe bump version? Unsure how it eventually makes its way into the wild...

scripts-available/CDB_UserTables.sql
```diff
@@ -6,34 +6,23 @@
 -- Currently accepted permissions are: 'public', 'private' or 'all'
 --
 CREATE OR REPLACE FUNCTION CDB_UserTables(perm text DEFAULT 'all')
-RETURNS SETOF information_schema.sql_identifier
+RETURNS SETOF name
```
rafatower: This shouldn't return a `SETOF name` cause that type is limited to 63 chars and schema + table_name can exceed that limit.

How about using `regclass` that has automatic casting to text types?

PR #84, opened 2015-06-08
Add distribution classifier

ohasselblad:
These functions classify a distribution of data into: A, J, U, S, F, L, where the distribution in the histogram resembles these letters. More information [here](http://druedin.com/2012/12/08/galtungs-ajus-system/).

cc @javisantana


ohasselblad:
@pramsey would you mind doing a code review of this?

scripts-available/CDB_DistType.sql
```diff
@@ -0,0 +1,138 @@
...
+        ELSE
+            type = 'S';
+        END IF;
+    END IF;
+
+    RETURN type;
+END;
+$$ language plpgsql IMMUTABLE;
+
+CREATE OR REPLACE FUNCTION CDB_CompareValues ( a numeric, b numeric, tolerance numeric ) RETURNS INT as $$
```
pramsey: I'd say this is better just inlined into the main function as a CASE statement.

scripts-available/CDB_DistType.sql
```diff
@@ -0,0 +1,138 @@
...
+        signature = CDB_DistributionType(ajus);
+    END IF;
+
+    RETURN signature;
+END;
+$$ language plpgsql IMMUTABLE;
+
+-- Classify data into AJUSFL
+
+CREATE OR REPLACE FUNCTION CDB_DistributionType ( in_array INT[] ) RETURNS text as $$
```
pramsey: For functions that are called as part of other functions but not really of use themselves, consider marking them as private by (a) using an _ as the start of the signature and (b) naming them into a hierarchy with their parent (ie, _CDB_DistTypeClassify)

scripts-available/CDB_DistType.sql
```diff
@@ -0,0 +1,138 @@
...
+-- 
+-- Future improvements:
+--    variable number of bins (7 is baked in right now)
+--    catch the number of items to ensure that the sample is large enough
+--
+-- Refs:
+--    1. width_bucket/histograms: http://tapoueh.org/blog/2014/02/21-PostgreSQL-histogram
+--    2. R implementation: https://github.com/cran/agrmt
+
+CREATE OR REPLACE FUNCTION CDB_DistType ( in_array NUMERIC[] ) RETURNS text as $$
```
pramsey: The thing that makes my most disquiet here is shoving the contents of an entire column into this function as an array. This is the very definition of "not scalable", though I know the pattern is all over the place in CDB. Realistically, for a lot of these functions, we should be adding cached metadata about distribution, extents, etc, to the system rather than passing full tables scans through functions like this. There is actually already a data histogram for every indexed numeric column in the system, in the stats table, though it might not exist immediately upon import (it requires an ANALYZE to be run on the table). @rambo

ohasselblad: @javisantana mentioned that I should use the histogram in pg_stats instead. That seems like it would solve this problem?

pramsey: In exchange for another problem, which is "what if there are no stats for the column currently and the process running the function doesn't have privs to run ANALYZE"? Which is a deeper, system-level problem for CartoDB. If we're going to start leaning on the stats table for user-facing functionality, we need to take control over ensuring it is kept up-to-date. (Also, I need to check that stats are gathered for non-indexed columns, not sure that they are.)

scripts-available/CDB_DistType.sql
ohasselblad: +1

scripts-available/CDB_DistType.sql
ohasselblad: +1

ohasselblad:
I'm going to merge this now so I can test it in production. Thanks @pramsey for the CR!

scripts-available/CDB_DistType.sql
javisantana: So maybe we should have an specialization for this function like CDB_DistType(table, column_name) ? In that case we could either use internal stats or calculate them

PR #87, opened 2015-06-23
Function deciding criteria for using a category column in a map

ohasselblad:
This function looks at a text column, breaks it into it's distinct categories, sorts by most common categories, and tells whether threshold % of the rows are in the first 10 or fewer major categories. Part of pecan cookies.

@pramsey would you mind doing a code review of this one? Apologies in advance about ingesting an entire column. Should it be written as an aggregate function instead?

cc @stuartlynn


pramsey:
What is this *for*? It looks like something that should be integrated somewhere else, or left for folks to build as needed (it's not *much* more than `SELECT Count(*), foo FROM bar GROUP BY foo` at heart). The fact that again it's taking in the contents of a whole table is again worrisome since probably it'll end up sprinkled here and there in multiple places. That's a code pattern we should be removing, not encouraging.

ohasselblad:
I needed this function to [score a column](https://github.com/CartoDB/cartodb-management/wiki/Pecan-Cookies-(A-la-instagram-maps)) that could be a category map. I'm open to putting it elsewhere, but need guidance on where to put it instead @javisantana 

I agree that taking the entire contents of the column is not a good practice. Should I be doing an aggregate function instead?

javisantana:
this is know what is the best visualization approach for a column. My proposal for this function is to use real stats for small tables and postgres stats for bigger ones but that's an implementation detail.

For the the interface should be more like:

```
CDB_DistinctMeasure(regclass, column_name, threshold)
```

so we could use stats. 


For the moment we will use it with small size tables (less than 5k rows)

pramsey:
If you do an aggregate it wouldn't be for *this* per se, but for something that covers over the larger task of binning and analyzing a histogram in one step.

rochoa:
:+1:

I'm not an expert here but I think it would be useful to have a param to pass the number of categories, but we can add that later as well.

So ... :shipit:

ohasselblad:
Thanks @rochoa and @pramsey 
I'll open a ticket about @pramsey's concerns and @javisantana's suggestions and flag other functions that have similar issues.

Issue #86, opened 2015-06-18
Long table names causing trouble


rafatower:
See https://github.com/CartoDB/cartodb/issues/3808

The problem is the 63 char limit for identifiers.
- `CDB_QueryTables` returns `name[]`, containing `"schema.table"`'s that can exceed 63 chars in total
- When trying to join that with `CDB_TablesMetadata`, `name`'s potentially do not fit in a `regclass`

We should probably modify `CDB_QueryTables` to return `(schema, table)` pairs

rafatower:
Some quick tests against 9.3:
```sql
-- String literal to regclass OK
SELECT (('public.bug3808_renamed_cpre_charter_to_save_our_countryside_comments_e')::regclass);

-- String literal to NAME to regclass ERROR
SELECT (('public.bug3808_renamed_cpre_charter_to_save_our_countryside_comments_e')::name::regclass);

-- String to name to text to regclass ERROR
SELECT (('public.bug3808_renamed_cpre_charter_to_save_our_countryside_comments_e')::name::text::regclass);
```
It will be hard to find a workaround for this without modifying `CDB_QueryTables`

rafatower:
Makes sense since `name` is limited to 64 bytes: http://www.postgresql.org/docs/9.3/static/datatype-character.html

The quickest thing to do is create another version of `CDB_QueryTables` with different return type.

rochoa:
With the help of @rafatower we just validated that using regclass[] as return type would fix some of the problems we have in https://github.com/CartoDB/Windshaft-cartodb/blob/2.5.0/lib/cartodb/api/query_tables_api.js#L36-L43 Find bellow a patch.

However changing the return type might break some uses of the function.

```patch
diff --git a/scripts-available/CDB_QueryTables.sql b/scripts-available/CDB_QueryTables.sql
index cd8b51b..c304553 100644
--- a/scripts-available/CDB_QueryTables.sql
+++ b/scripts-available/CDB_QueryTables.sql
@@ -3,11 +3,11 @@
 -- Requires PostgreSQL 9.x+
 --
 CREATE OR REPLACE FUNCTION CDB_QueryTables(query text)
-RETURNS name[]
+RETURNS REGCLASS[]
 AS $$
 DECLARE
   exp XML;
-  tables NAME[];
+  tables REGCLASS[];
   rec RECORD;
   rec2 RECORD;
 BEGIN
@@ -41,11 +41,11 @@ BEGIN
[...]

PR #88, opened 2015-06-24
Add a new function CDB_QueryTablesText #86

rafatower:
The return values of it can be safely used when len(schema.table_name)
exceeds the 63 char limit of the postgres type `name`.

@rochoa please review

### DIFF
--- a/scripts-available/CDB_QueryTables.sql
+++ b/scripts-available/CDB_QueryTables.sql
@@ -2,12 +2,12 @@
 --
 -- Requires PostgreSQL 9.x+
 --
-CREATE OR REPLACE FUNCTION CDB_QueryTablesRegclass(query text)
-RETURNS regclass[]
+CREATE OR REPLACE FUNCTION CDB_QueryTablesText(query text)
+RETURNS text[]
 AS $$
 DECLARE
   exp XML;
-  tables regclass[];
+  tables text[];
   rec RECORD;
   rec2 RECORD;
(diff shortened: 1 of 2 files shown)


rochoa:
I'm happy with the changes but the tests are not that happy :smiling_imp:

rafatower:
yeah, I'm trying to figure out what happens

@rochoa one question about that: is the schema name absolutely necessary? because I've seen that pg only returns the schema name from the regclass when there's ambiguity (two tables with same name in different schemas).

rochoa:
It's not critical as far as it's consistent. Feel free to modify the tests and add a test to validate your *ambiguity* theory and I'll be more than happy! :smile: 

**Important**: when releasing this we will have to prune any cached resource relying on having always schema.table_name format.

rafatower:
regclass is "search_path aware", it does not return the schema if the referenced table is found in the search path. Internally it is equivalent to using OID's with text autocasting powers. Is that enough?

In that case I can write tests for it and modify existing ones accordingly.

Otherwise I can peek into catalog tables to return the schema and the table_name but then instead of a `CDB_QueryTablesRegclass` we will have `CDB_QueryTablesText` to overcome 63 chars limitation and keep the other for compatibility based on it.

rafatower:
@rochoa please review again. I made sure it is backwards compatible.

rochoa:
Fine by me!

PR #92, opened 2015-06-30
Fix versioning by creating a 0.8.0

rafatower:
Fix versioning by creating a new major version since it contains new
features. Keep version 0.7.4 that should've never existed and provide an
upgrade path for the new version.

@rochoa can you please review?


rochoa:
:+1: thanks for taking care of versioning :heart:

PR #99, opened 2015-07-27
added tests for usertables to check private tables are not returned

javisantana:


javisantana:
cc @Kartones

Kartones:
looks ok :+1:

PR #107, opened 2015-08-11
Do not create timestamp columns/triggers on cartodbfy

rafatower:
This solves the issue when timestamp triggers are triggered.

@pramsey can you please review?


test/CDB_CartodbfyTableTest.sql
```diff
@@ -112,15 +102,6 @@ BEGIN
...
 
-  -- Check null constraint on cartodb_id, created_at, updated_at
-  SELECT count(*) FROM pg_attribute a, pg_class c WHERE c.oid = tabname::oid
-    AND a.attrelid = c.oid AND NOT a.attisdropped AND a.attname in
-      ( 'cartodb_id', 'created_at', 'updated_at' )
-    AND NOT a.attnotnull INTO strict tmp;
-  IF tmp > 0 THEN
-      RAISE EXCEPTION 'cartodb_id or created_at or updated_at are missing not-null constraint';
-  END IF;
-
```
pramsey: You've lost the not-NULL test on cartodb_id here, not sure if this table test function is used anywhere though.

test/CDB_QuotaTest_expect
```diff
@@ -9,7 +9,7 @@ INSERT 0 2048
 INSERT 0 2048
 INSERT 0 2048
 2
-ERROR:  Quota exceeded by 567.998046875KB
+ERROR:  Quota exceeded by 519.998046875KB
 0
```
pramsey: I feel like I've seen this swing back and forth a bit... I have a feeling there's a platform dependence in the code somewhere.

rafatower: Well, I think it is normal since the table has not the timestamp columns it takes less storage space.

test/CDB_CartodbfyTableTest.sql
rafatower: good catch, I'll recover the not null check for cartodb_id

PR #109, opened 2015-08-13
Replace CDB_CartodbfyTable by new CartodbfyTable2

rafatower:
In this PR:
- Delete old CDB_CartodbfyTable code
- Delete auxiliary functions no longer used
- Modify the new CDB_CartodbfyTable signature to be backwards
  compatible.

Also, there are some tests failing. I copy below the relevant output:
```
***************
*** 51,58 ****
  FROM CDB_TableMetadata WHERE tabname = 'c.t3'::regclass;
   tabname | age 
  ---------+-----
!  c.t3    |   0
! (1 row)
  
  SET SESSION AUTHORIZATION 'cartodb_postgresql_unpriv_user';
  -- Table with cartodb_id field, see
--- 51,57 ----
  FROM CDB_TableMetadata WHERE tabname = 'c.t3'::regclass;
   tabname | age 
  ---------+-----
! (0 rows)
  
  SET SESSION AUTHORIZATION 'cartodb_postgresql_unpriv_user';
  -- Table with cartodb_id field, see
```
(Tables not getting registered as expected in cdb_tablemetadata. See below)

```
--- 126,132 ----
  (1 row)
  
  alter table c.t3 drop column the_geom_webmercator;
! ERROR:  must be owner of relation t3
  RESET SESSION AUTHORIZATION;
  select
   tabname::text,
```
(some issue with either ownership or schema, which is weird because in the original cartodbfy `source_schema == dest_schema`)

```
single non-geometrical column cartodbfied fine
DROP TABLE
SELECT 1
ERROR:  syntax error at or near "8225081"
LINE 1: INSERT INTO 8225081(the_geom) values ( CDB_LatLng(2,1) ) RET...
```
(weird, I think that is the `CDB_CartodbfyTableCheck`getting the original oid in `tabname::text`)

```
! ERROR:  duplicate key value violates unique constraint "big_pkey"
! DETAIL:  Key (cartodb_id)=(1) already exists.
```
(minor issue, this has to do with the ordering of the columns).

@pramsey can you please review and propose some fixes?


pramsey:
OK, just starting from the first test failures (there are so many!)

A key problem is the DDL triggers, which are set up to Cartodbfy newly created tables automatically. They are also built with the assumption that Cartodbfy works in-place, which is no longer true. So these errors where we're getting numbers instead of table names is because a regclass value has been inserted and then the table has been dropped (as Cartodbfy2 drops the original table after creating the properly formatted one).

For extra scariness, there's even a comment in the DDL triggers, to the effect that 
```
  -- CDB_CartodbfyTable must not create tables, or infinite loop will happen
```
But further reading shows that someone is turning off the event triggers before running Cartodbfy and then turning them back on again after (I should git blame to figure out when that happened).

Probably the straightest fix is to have Cartodbfy2 return the Oid the cartodbfy'ed table, so that downstream functions that want to use it have that option...

pramsey:
Chipping away at these, down to mostly just differences in log messages. There are still some functional differences though. The big thing continues to be the fact that the new Cartodbfy rewrites a new table. The current issue I'm wrestling with is the event trigger on create table is created as a security definer
```
CREATE OR REPLACE FUNCTION cartodb.cdb_handle_create_table ()
RETURNS event_trigger SECURITY DEFINER LANGUAGE plpgsql AS $$
```
which means that the Cartodbfy() function that runs w/i it is also called with superuser powers, which in turn means that the new table it creates is owned by the superuser and then everything falls apart, because it should be owned by the user that created it.

pramsey:
The commit message mostly says it all. In a change this big, there will be a few behaviour changes. I've fixed up the tests enough that we can now actually see the changes in behaviour in the test logs (rather than the tests just crashing and burning)
``` 
! ERROR:  type of the_geom_webmercator in geometry_columns is POINT, expected GEOMETRY
```
So, the old version coerced all geometry columns to the generic geometry type, and the new one is letting type constraints remain. That could be changed.
```
  text cartodb_id cartodbfied fine
! 5
```
In this case the test table had a text column `cartodb_id` and the code preserved it. The new code doesn't have as much leeway to check the results of things like that (without spending a lot of CPU time we are trying not to spend) and just creates a fresh `cartodb_id` if the column is not an integer type.
```
! ERROR:  column "_cartodb_id0" does not exist
! LINE 1: select cartodb_id,_cartodb_id0 FROM t;
```
In this case the linting of column names is slightly different when avoiding name collisions.
```
The other errors of note are the ones in the ddl triggers tests, where we are again being tripped up by the new cartodbfy writing a new table, which causes old regclass references to go stale. The `CDB_TableMetadata` table contains regclass entries, but when a column alter happens and Cartodbfy gets triggered, a new table gets written... uh oh, now the entry in `CDB_TableMetadata` no longer matches the oid of the table. Game over.

I've fixed it in the triggers, but I think all the mucking with `CDB_TableMetadata` should happen inside Cartodbfy itself, so we can paper over the regclass change problems a little more.

**Comment by rafatower** [comment]
@pramsey thank you very much for the fixes and insights. I'll continue from there

@javisantana the ddl triggers are holding us back and maybe we should consider deprecating that part of the extension.

**Comment by javisantana** [comment]
@rafatower we should consider start to use them, we need a better way to sync when a table is created/dropped, right now we are using a hack from the editor

What is the problem with ddl triggers exactly?

**Comment by rafatower** [comment]
@javisantana okay, I did not get it from the discussion in tech, my bad.

The problem basically is that they were designed in such a way they do not get along with `_CDB_Rewrite_Table()`. Paul work-arounded some of the problems with that but as far as I understand there are some remaining to get ddl triggers actually working, for instance and at least when a column alter happens (see Paul's comment above, oid in table metadata getting out of sync).

**Comment by rafatower** [comment]
As per a conversation with Javi:
- The ddl triggers are to be used but not for cartodbfy but just notify the editor of ddl changes and invalidations
- They are meant to notify but not to modify
- We can remove the triggering of cartodbfy process from ddl triggers

**Comment by rafatower** [comment]
Oookay! I think we can go ahead with these changes:
- I changed the test to allow for GEOMETRY and POINT as valid geometries
- I recovered the check_prerequisites from the trash
- I disabled the tests that expected cartodb_id text column to be casted. We will have to rely on ogr2ogr on doing its duty correctly on that respect.

**Comment by pramsey** [comment]
Despite the comment in the code that using a CREATE inside Cartodbfy could cause an infinite loop (CREATE calls event trigger which calls Cartodbfy which calls CREATE which...) the actual trigger implementator ended up doing a hack of 
```
  PERFORM cartodb.cdb_disable_ddl_hooks();
 
   -- CDB_CartodbfyTable must not create tables, or infinite loop will happen
  PERFORM cartodb.CDB_CartodbfyTable(rel.nspname, event_info.relation);
 
  PERFORM cartodb.cdb_enable_ddl_hooks();
```
thereby avoiding the infinite loop. So don't drop the event trigger thing on behalf of that. I do think that cartodbfying *every* table *all* the time might be a bridge too far, we'd be better off slowly making the editor able to handle more and more tables, regardless of column naming convention ("you got a primary key, and a geometry(4326)? I'll map you!")

**Comment by pramsey** [comment]
Oh, the other thing, for all the trouble we had cleaning up the DDL trigger tests to get them working: DDL triggers aren't running in production, right? I mean, I can create a table in production, it doesn't get automatically Cartodbfy'ed. It's just a bunch of code with a bunch of tests that currently isn't being used. We pass the tests again, for what that's worth.

**Comment by rafatower** [comment]
@pramsey: correct

**Comment by pramsey** [comment]
All the pain on these tests should raise a flag for implementation, because any code that does this:
```
foo = 'mytable'::regclass;
cartodbfy(foo);
dosomethingwith(foo);
```
will be wrong, as the regclass `foo` will no longer match `'mytable'::regclass` after the cartodbfy, thanks to table re-writing. Beware these patterns! the table retains its name through the cartodbfy, but the oid (probably) changes

rafatower:
Ok, I'll give it a shot, at least to prepare cartodbfy functions to return the new oid.
(aside, yeah, it raises tons of red flags :smile:)

rafatower:
@pramsey please review and if you're ok I think it is time to merge this back into `new_cartodbfy`, update with master and resume integration testing with the updated branch.

Have a nice weekend!

pramsey:
OK, continue work/discussion back on #78

PR #110, opened 2015-08-13
Fixes CDB_UserDataSize failing due `ERROR: relation "*" does not exist.`

### LINKED ISSUE #108: CDB_UserDataSize is not safe, might raise error
It might raise `ERROR: relation "*" does not exist`.

[`pg_total_relation_size` might run against an nonexistent relation](https://github.com/CartoDB/cartodb-postgresql/blob/master/scripts-available/CDB_Quota.sql#L27) the table list is [previously populated in a CTE](https://github.com/CartoDB/cartodb-postgresql/blob/master/scripts-available/CDB_Quota.sql#L12-L18) and between list generation and consumption it's possible some of the relations get deleted.

A safer version is required, handling the nonexistent relation case.

I've created a simpler snippet so the scenario is easier to reproduce:

```sql
DROP TABLE IF EXISTS public.foo;
CREATE table public.foo (a text);

CREATE OR REPLACE FUNCTION drop_table()
RETURNS bigint AS $$
BEGIN
  DROP TABLE public.foo;
  return 0;
END;
$$
LANGUAGE 'plpgsql' VOLATILE;

WITH user_tables AS (
  SELECT table_name FROM information_schema.tables
  WHERE table_catalog = current_database() AND table_schema = 'public'
    AND table_name != 'spatial_ref_sys'
    AND table_name != 'cdb_tablemetadata'
    AND table_type = 'BASE TABLE'
),
dropping_table AS (
  SELECT drop_table() as wadus
)
SELECT table_name, pg_total_relation_size('"' || 'public' || '"."' || table_name || '"') table_size
FROM user_tables, dropping_table
```

p.s. just by removing the `dropping_table` in the last line from the `FROM` clause it will work as, I guess, the query planner knows it doesn't need to execute that CTE.

rochoa:
Adds new _CDB_total_relation_size function that handles nonexistent
tables and does fallback to size=0.

That function could be used to cache total relation size or query another
table view with a cached total relation size.

Fixes #108

cc @javisantana @Kartones


scripts-available/CDB_Quota.sql
```diff
@@ -1,3 +1,22 @@
...
+$$
+BEGIN
+  IF EXISTS (
+      SELECT 1 FROM information_schema.tables
+      WHERE table_catalog = current_database()
+        AND table_schema = _schema_name
+        AND table_name = _table_name
+  )
+  THEN
+    RETURN pg_total_relation_size(format('"%s"."%s"', _schema_name, _table_name));
```
javisantana: wouldn't be possible that the table is removed in between the "if" and the pg_total_relation_size call ?

Kartones: If Javi is right, maybe doing like cartodbfication and [relying on exceptions](https://github.com/CartoDB/cartodb-postgresql/blob/master/scripts-available/CDB_CartodbfyTable.sql#L81) could be safer?

In any case this call should be really quick no?

rochoa: I'm not fully sure. As an alternative I had:

```sql
CREATE OR REPLACE FUNCTION cartodb._CDB_total_relation_size(_schema_name TEXT, _table_name TEXT)
RETURNS bigint AS
$$
DECLARE relation_size bigint;
BEGIN
  relation_size := 0;
  SELECT pg_total_relation_size(format('"%s"."%s"', _schema_name, _table_name)) INTO relation_size WHERE EXISTS (
      SELECT 1 FROM information_schema.tables
      WHERE table_catalog = current_database()
        AND table_schema = _schema_name
        AND table_name = _table_name
  );
  return relation_size;
END;
$$
LANGUAGE 'plpgsql' VOLATILE;
```

But it's kind of the same. The alternative is use a EXCEPTION catch but I was trying to avoid it.

cc @pramsey any thoughts?

pramsey: None. I like your better.

rochoa: But... as @javisantana ia asking: is it safe? :smiling_imp:

pramsey: Yes, although so is the other one, as it too is run within a single transaction block (all pl/pgsql functions are unitary transaction blocks, so state won't change on things in them unless you change that state yourself).

javisantana: So then we have the race condition in the previous query?

pramsey: You might have a race in the callers of the functions, but not in the functions themselves... if someone calls CDB_UserDataSize on a table that disappeared before they called it.

rochoa: Then ... probably I'm doing something really wrong here: https://github.com/CartoDB/cartodb-postgresql/issues/108

rochoa:
So I guess we can conclude the changes proposed in this PR are fine and I can merge it. Ok?

pramsey:
unknown

rochoa:
Thanks a lot! :+1:

PR #112, opened 2015-08-13
Review test to validate permissions in public tables

rochoa:
This activates parts of the already existent test that validates granting `publicuser` to a user allows to select from other users' public tables.

Ref: https://github.com/CartoDB/cartodb/pull/5021


rochoa:
No need to bump version as it only includes test changes.

PR #119, opened 2015-08-20
Consider the_geom in 4326 if no SRID #117

### LINKED ISSUE #117: Import dataset fail in staging
Importing a dataset in staging gives an upload error that doesn't happen in production.

Follow these steps to reproduce it:

1. Download [this dataset](https://mariacheca.cartodb.com/tables/public.meteorite_landings/public/map) as csv.
2. Try to upload it from the dataset uploader with version 0.9.1 of the extension **disabling data type guessing**.
3. Fails, returning this:
```
2015-08-20 09:11:03 UTC: Exception: PG::Error: ERROR:  column "the_geom" of relation "public_1" does not exist
CONTEXT:  SQL statement "CREATE trigger update_the_geom_webmercator_trigger BEFORE INSERT OR UPDATE OF the_geom ON public_1 FOR EACH ROW EXECUTE PROCEDURE cartodb._CDB_upd

app/models/table.rb:1176:in `block in cartodbfy'
```

rafatower:
This fixes #117 

@pramsey & @javisantana please review

### DIFF
--- a/scripts-available/CDB_CartodbfyTable.sql
+++ b/scripts-available/CDB_CartodbfyTable.sql
@@ -925,12 +925,15 @@ BEGIN
       AND t.typname = 'geometry'
       AND a.attnum > 0
       AND NOT a.attisdropped
-      AND (postgis_typmod_srid(a.atttypmod) > 0 OR srid.srid > 0)
+      AND ((postgis_typmod_srid(a.atttypmod) > 0 OR srid.srid > 0) OR (a.attname = const.geomcol))
       ORDER BY a.attnum
       LIMIT 1;
     
       IF FOUND THEN
         sql := sql || geom_transform_sql;
+      ELSE
+        RAISE NOTICE 'rtorre: spetial tricky one';
+	geom_column_source := '';
       END IF;
 
     END IF;


pramsey:
How does it fix it? If the SRID = 0, then the ST_Transform() should fail further down the line, no?

rafatower:
Because actually the_geom came from EWKT. This is roughly what is happening:

```sql
-- Table as it comes from the csv file
SELECT ST_SetSRID(st_point(1,2),4326)::text as the_geom INTO bug117;

# \d bug117
    Table "public.bug117"
  Column  | Type | Modifiers 
----------+------+-----------
 the_geom | text | 

-- Rails doing a bit of magic, based on name of the columns
ALTER TABLE bug117 ALTER the_geom TYPE geometry USING the_geom::geometry;

-- SRID set to 0
SELECT Find_SRID('public', 'bug117', 'the_geom');
 find_srid 
-----------
         0

-- Then the st_transform in CDB_CartodbfyTable
SELECT ST_AsText(ST_Transform(the_geom, 4326)) FROM bug117;
 st_astext  
------------
 POINT(1 2)
(1 row)
```

scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -925,12 +925,15 @@ BEGIN
...
       AND NOT a.attisdropped
-      AND (postgis_typmod_srid(a.atttypmod) > 0 OR srid.srid > 0)
+      AND ((postgis_typmod_srid(a.atttypmod) > 0 OR srid.srid > 0) OR (a.attname = const.geomcol))
       ORDER BY a.attnum
       LIMIT 1;
     
       IF FOUND THEN
         sql := sql || geom_transform_sql;
+      ELSE
+        RAISE NOTICE 'rtorre: spetial tricky one';
```
javisantana: do you want to keep this? :D

rafatower: I can change the string to not point to myself but yeah, having that trace can help me debug that other corner case if it appears :)

I'm thinking on another possible solution, just forcing the column types as it used to work.

rafatower:
This doesn't look like the best solution, closing this PR..

PR #124, opened 2015-08-26
Add kurtosis and skewness

ohasselblad:
Adding these measures as they could be used in pecan analysis

To do: still need to add tests

### DIFF
--- a/scripts-available/CDB_Stats.sql
+++ b/scripts-available/CDB_Stats.sql
@@ -1,0 +1,33 @@
+--
+-- Calculate the Pearson kurtosis of the input data
+--
+-- @param in_array A numeric array of numbers to determine the best
+--                   to determine the bin boundary
+--
+-- @param breaks The number of bins you want to find.
+--  
+--
+-- Returns: upper edges of bins
+-- 
+--
+
+CREATE OR REPLACE FUNCTION CDB_Kurtosis ( in_array NUMERIC[] ) RETURNS NUMERIC as $$
+DECLARE 
+    a numeric;
+    c numeric;
+    s numeric;
+    k numeric;
+BEGIN
+    SELECT AVG(e), COUNT(e)::numeric, stddev(e) INTO a, c, s FROM ( SELECT unnest(in_array) e ) x;
+
+    RAISE NOTICE 'avg: %, cnt: %, std: %', a, c, s;
+
+    EXECUTE '
+        SELECT sum(power($1 - e,4)) / ( $2 * power($3, 4))
+        FROM (SELECT unnest($4) e ) x'
+    INTO k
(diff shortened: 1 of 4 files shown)


rochoa:
Hey @ohasselblad, it would be nice if you add some tests for these functions.

ohasselblad:
ok @rochoa added them, but the functions aren't being found in the test. It must be in my naming convention. Should I not be bundling in functions that do two different things in one file?

rochoa:
I did a couple of changes:
 - fix the symlink: that was preventing to make the functions available during tests
 - raise the min messages to not validate notices

test/CDB_StatsTest.sql
```diff
@@ -0,0 +1,16 @@
+-- continuous uniform distribution has kurtosis = -6/5, skewness = 0.0
+-- http://mathworld.wolfram.com/UniformDistribution.html
+set client_min_messages to ERROR;
+
+With dist As (
+  SELECT random()::numeric As val
+  FROM generate_series(1,5000000) t
```
rochoa: Can we test with a lower range? Testing with such a big one means the test is very slow.

What is the minimum range we can use to validate these functions?

scripts-available/CDB_Stats.sql
```diff
@@ -0,0 +1,51 @@
...
+CREATE OR REPLACE FUNCTION CDB_Kurtosis ( in_array NUMERIC[] ) RETURNS NUMERIC as $$
+DECLARE
+    a numeric;
+    c numeric;
+    s numeric;
+    k numeric;
+BEGIN
+    SELECT AVG(e), COUNT(e)::numeric, stddev(e) INTO a, c, s FROM ( SELECT unnest(in_array) e ) x;
+
+    RAISE NOTICE 'avg: %, cnt: %, std: %', a, c, s;
```
rochoa: Can we use debug instead?

scripts-available/CDB_Stats.sql
```diff
@@ -0,0 +1,51 @@
...
+CREATE OR REPLACE FUNCTION CDB_Skewness ( in_array NUMERIC[] ) RETURNS NUMERIC as $$
+DECLARE
+    a numeric;
+    c numeric;
+    s numeric;
+    sk numeric;
+BEGIN
+    SELECT AVG(e), COUNT(e)::numeric, stddev(e) INTO a, c, s FROM ( SELECT unnest(in_array) e ) x;
+
+    RAISE NOTICE 'avg: %, cnt: %, std: %', a, c, s;
```
rochoa: Same about using debug.

ohasselblad:
:+1: thanks for all the input, @rochoa :)

Issue #123, opened 2015-08-25
Failure importing from ogr2ogr with -overwrite


rafatower:
Scenario to reproduce:
1. Import a table called `country_lat_long`
2. Rename it to `country_lat_long_whatever`
3. Run this:
```
ogr2ogr2 \
  --config CARTODB_API_URL https://myuser.cartodb.com/api/v2/sql \
  --config CARTODB_API_KEY 1234 \
  -s_srs EPSG:4326 \
  -t_srs EPSG:4326 \
  -f CartoDB "CartoDB:myuser" \
  -oo X_POSSIBLE_NAMES=lon\* \
  -oo Y_POSSIBLE_NAMES=lat\* \
  -overwrite \
  -dim 2 \
  country_lat_long.csv
```

Expected: the table is imported correctly

Actual:
```
ERROR 1: HTTP error code : 400
ERROR 1: Error returned by server : CDB(_CDB_Add_Indexes:42P07:relation "country_lat_long_the_geom_gix" already exists): CREATE INDEX country_lat_long_the_geom_gix ON country_lat_long USING GIST (the_geom)
ERROR 1: Unable to write feature 1 from layer country_lat_long.

ERROR 1: Terminating translation prematurely after failed
translation of layer country_lat_long (use -skipfailures to skip errors)
```

rafatower:
also seen in cartodb repo tests:
```
09:09:16   1) Table table setups receives a name change if table visualization name changed
09:09:16      Failure/Error: table = create_table(name: 'bogus_name', user_id: $user_1.id)
09:09:16      Sequel::DatabaseError:
09:09:16        PG::Error: ERROR:  CDB(_CDB_Add_Indexes:42P07:relation "bogus_name_the_geom_gix" already exists): CREATE INDEX bogus_name_the_geom_gix ON bogus_name USING GIST (the_geom)
09:09:16        CONTEXT:  SQL statement "SELECT _CDB_SQL(sql, '_CDB_Add_Indexes')"
09:09:16        PL/pgSQL function _cdb_add_indexes(regclass) line 89 at PERFORM
09:09:16        SQL statement "SELECT _CDB_Add_Indexes(destoid)"
09:09:16        PL/pgSQL function cdb_cartodbfytable(text,regclass) line 57 at PERFORM
09:09:16      # ./app/models/table.rb:1181:in `block (2 levels) in cartodbfy'
...
```

We're doing something wrong with indexes.

rafatower:
[...]

PR #127, opened 2015-08-27
Fix for index generation when renaming table #123

rafatower:
Just let the DB decide the name of the index, as it was done in the previous version.

@rochoa or @pramsey please kindly review.

### DIFF
--- a/scripts-available/CDB_CartodbfyTable.sql
+++ b/scripts-available/CDB_CartodbfyTable.sql
@@ -1234,7 +1234,7 @@ BEGIN
     AND c.oid = reloid
     AND am.amname != 'gist'
   LOOP
-    sql := Format('CREATE INDEX %s_%s_gix ON %s USING GIST (%s)', relname, rec.attname, reloid::text, rec.attname);
+    sql := Format('CREATE INDEX ON %s USING GIST (%s)', reloid::text, rec.attname);
     PERFORM _CDB_SQL(sql, '_CDB_Add_Indexes');
   END LOOP;
     
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+0.9.4 (2015-08-28)
+------------------
+* Fixed issue with indices when renaming tables [#123](https://github.com/CartoDB/cartodb-postgresql/issues/123)
+
 0.9.3 (2015-08-27)
 ------------------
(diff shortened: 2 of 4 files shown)


pramsey:
:+1:

scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -1234,7 +1234,7 @@ BEGIN
     AND c.oid = reloid
     AND am.amname != 'gist'
   LOOP
-    sql := Format('CREATE INDEX %s_%s_gix ON %s USING GIST (%s)', relname, rec.attname, reloid::text, rec.attname);
+    sql := Format('CREATE INDEX ON %s USING GIST (%s)', reloid::text, rec.attname);
```
rochoa: What happens with old indexes? Do they stay in place also?

pramsey: It’s happening w/i the context of a loop that is finding only geometry columns that don’t already have gist indexes. So as long as the table rename retained the reference to the previous indexes, no extra indexes will get created.

rochoa: :-) Thanks for the explanation. :ship: it!

On Thu, Aug 27, 2015 at 4:48 PM, Paul Ramsey <notifications@github.com>
wrote:

> In scripts-available/CDB_CartodbfyTable.sql
> <https://github.com/CartoDB/cartodb-postgresql/pull/127#discussion_r38103736>
> :
>
> > @@ -1234,7 +1234,7 @@ BEGIN
> >      AND c.oid = reloid
> >      AND am.amname != 'gist'
> >    LOOP
> > -    sql := Format('CREATE INDEX %s_%s_gix ON %s USING GIST (%s)', relname, rec.attname, reloid::text, rec.attname);
> > +    sql := Format('CREATE INDEX ON %s USING GIST (%s)', reloid::text, rec.attname);
>
> It’s happening w/i the context of a loop that is finding only geometry
> columns that don’t already have gist indexes. So as long as the table
> rename retained the reference to the previous indexes, no extra indexes
> will get created.
>
> —
> <https://github.com/CartoDB/cartodb-postgresql/pull/127/files#r38103736>.
>



-- 
Raul Ochoa
rochoa@cartodb.com

PR #131, opened 2015-09-02
Column regclass

### LINKED ISSUE #122: CDB_ColumnNames should honor regclass
[CDB_ColumnNames current implementation](https://github.com/CartoDB/cartodb-postgresql/blob/0.9.1/scripts-available/CDB_ColumnNames.sql) returns columns for any table matching the name.

When having multiple tables with same name but in a different schema it will report the column name from all the tables.

`regclass` should be considered when providing qualified table names.

### LINKED ISSUE #130: CDB_ColumnType should honor regclass
This is similar to #122

rochoa:
Uses schema and table name from regclass to avoid query tables with same name in different schemas.

Fixes #122 and fixes #130.

### DIFF
--- a/scripts-available/CDB_ColumnNames.sql
+++ b/scripts-available/CDB_ColumnNames.sql
@@ -7,7 +7,7 @@ AS $$
       FROM information_schema.columns c, pg_class _tn, pg_namespace _sn
       WHERE table_name = _tn.relname
         AND table_schema = _sn.nspname
-        AND _tn.oid = $1::regclass::oid
+        AND _tn.oid = $1::oid
         AND _sn.oid = _tn.relnamespace;
          
 $$ LANGUAGE SQL;
--- a/scripts-available/CDB_ColumnNames.sql
+++ b/scripts-available/CDB_ColumnNames.sql
@@ -3,11 +3,12 @@ CREATE OR REPLACE FUNCTION CDB_ColumnNames(REGCLASS)
 RETURNS SETOF information_schema.sql_identifier
 AS $$
 
-    SELECT column_name 
-      FROM information_schema.columns 
-      WHERE
-        table_name IN (SELECT CDB_UserTables())
-        AND table_name = '' || $1 || '';
+    SELECT c.column_name
+      FROM information_schema.columns c, pg_class _tn, pg_namespace _sn
+      WHERE table_name = _tn.relname
+        AND table_schema = _sn.nspname
+        AND _tn.oid = $1::regclass::oid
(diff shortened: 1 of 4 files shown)


rochoa:
@rafatower, what do you think?

scripts-available/CDB_ColumnNames.sql
```diff
@@ -3,11 +3,12 @@ CREATE OR REPLACE FUNCTION CDB_ColumnNames(REGCLASS)
...
-    SELECT column_name 
-      FROM information_schema.columns 
-      WHERE
-        table_name IN (SELECT CDB_UserTables())
-        AND table_name = '' || $1 || '';
+    SELECT c.column_name
+      FROM information_schema.columns c, pg_class _tn, pg_namespace _sn
+      WHERE table_name = _tn.relname
+        AND table_schema = _sn.nspname
+        AND _tn.oid = $1::regclass::oid
```
rafatower: is the casting to regclass needed?

scripts-available/CDB_ColumnNames.sql
```diff
@@ -3,11 +3,12 @@ CREATE OR REPLACE FUNCTION CDB_ColumnNames(REGCLASS)
 RETURNS SETOF information_schema.sql_identifier
 AS $$
 
-    SELECT column_name 
-      FROM information_schema.columns 
-      WHERE
-        table_name IN (SELECT CDB_UserTables())
```
rafatower: this line seems to be a safeguard against peeking in tables in which you don't have permissions, just wondering why it was added...

scripts-available/CDB_ColumnNames.sql
rochoa: No, it isn't. Removing it.

rafatower:
:+1:

scripts-available/CDB_ColumnNames.sql
rochoa: That's a good point.

I guess the intention was to filter by user tables, the problem is that CDB_UserTables returns `setof name` so when there was more than one table with same name in different schemas it was selecting columns from all of them.

I'm adding a couple of asserts in the test to validate that.

rochoa: See https://github.com/CartoDB/cartodb-postgresql/pull/131/files#diff-b7c395666b6f561e7618f21f8542d0cdR353

rafatower: then I prefer this version :smile:

PR #134, opened 2015-09-03
Quote schema and table names returned by CDB_QueryTables

rochoa:
Use `quote_ident` to quote schema and table names when necessary.


rochoa:
@rafatower, @pramsey 

What do you think? Any other scenario we may consider?

thanks

rafatower:
:+1: my concern would be with usages from the editor where a double escaping might be possible. But the change is correct and we'll have to deal with it in the editor in case that happens.

pramsey:
+1

Issue #138, opened 2015-09-07
Table cartodbfy'ed without default seq value


rafatower:
Steps to reproduce:
- Create an empty table:
```sql
CREATE TABLE bug_empty_table_no_seq (
    cartodb_id integer,
    the_geom geometry(Geometry,4326),
    the_geom_webmercator geometry(Geometry,3857),
    name text,
    description text
);
```
- Cartodbfy the table:
```sql
SELECT CDB_CartodbfyTable('bug_empty_table_no_seq');
```
- Try to insert an empty row:
```sql
INSERT INTO bug_empty_table_no_seq DEFAULT VALUES;
```

- Expected: the row is inserted without problem
- Actual:
```
ERROR:  null value in column "cartodb_id" violates not-null constraint
DETAIL:  Failing row contains (null, null, null, null, null).
```

The problem is that the table is created without the default sequence value. That breaks the editor and possibly other scenarios.

See https://github.com/CartoDB/cartodb/issues/5392

rafatower:
_CDB_Has_Usable_Primary_ID does not take into account if there's a valid sequence there. I think I got a fix for that particular case, but more profound changes might be required.

PR #140, opened 2015-09-08
138 fix cartodbfy no default seq value2

rafatower:
What's in this PR:
- added a small improvement to `test.sh` in 8c20b04f51fcd1aca1c97266d193c49b230b5e49. This is kind of unrelated to the fix when I tried to write a new test using it but finally could not without breaking things (mostly related to schemas).
- added a test for the failing scenario in bdc6b733df2d29a45097f0a9ac52744c2108661d
- `useable_key` has been renamed to `unique_key` within `_CDB_Has_Usable_Primary_ID`.
- added a bool var `has_sequence` to track whether the `cartodb_id` column has a sequence associated to it. Also modified the return value of `_CDB_Has_Usable_Primary_ID` to take it into account

See #138 for more information.

@rochoa @pramsey please review


rafatower:
PS: this is a rebase of [another branch](https://github.com/CartoDB/cartodb-postgresql/compare/138-fix-cartodbfy-no-default-seq-value?expand=1) where I did many experiments to achieve that solution. As in other PR's, this tries to fix the problem at hand while minimizing the impact. Nevertheless we will need to clarify and document the expected behavior of cartodbfy in different scenarios and rework the code.

rafatower:
Sorry, this doesn't work because it does not respect ID's when exporting and re-importing. So unless someone has a better solution, I will use the legacy [_CDB_create_cartodb_id_column](https://github.com/CartoDB/cartodb-postgresql/blob/45383d7c8a47c9b53cde0c118639d4399708c953/scripts-available/CDB_CartodbfyTable.sql#L55)

PR #143, opened 2015-09-09
Readd update_updated_at function (still used by old tables)

zenitraM:
The _CDB_update_updated_at() function was no longer added to tables since #107, but it is attached with a trigger to old tables, causing a pg_dump and further pg_restore of a database of this kind to fail as the current version (0.10.0) does not have this function.


zenitraM:
Could you take a look @rochoa or @rafatower? :bow: :smile:

rochoa:
OK from my side :+1: 

I relaunched the push travis build as it failed due an inconsistent test.

rafatower:
:+1:

PR #146, opened 2015-09-09
138 fix cartodbfy no default seq value3

rafatower:
@pramsey please review.

@rochoa there's a minor improvement in the "new test framework" here 6e4a5b5635f8e8cc6a5bc0e0e81f5ca1239b0ec8 (finally not used in these tests because of the schema issues with it)


scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -894,18 +923,20 @@ BEGIN
...
+    ELSE
+
+      RAISE DEBUG 'CDB(_CDB_Rewrite_Table): perfect table in the perfect place';
 
-  -- Don't move anything, just make sure our destination information is set right
-  ELSIF has_usable_primary_key AND gc.has_usable_geoms AND destschema = relschema THEN
+    END IF;
 
-    RAISE DEBUG 'CDB(_CDB_Rewrite_Table): perfect table in the perfect place';
     RETURN true;
```
rafatower: always return here if there's no need to rewrite

scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -933,11 +964,7 @@ BEGIN
   sql := Format('CREATE TABLE %s AS SELECT ', copyname);
 
   -- Add cartodb ID!
-  IF has_usable_primary_key THEN
-    sql := sql || const.pkey;
-  ELSE
-    sql := sql || 'nextval(''' || destseq || ''') AS ' || const.pkey;
-  END IF;
+  sql := sql || 'nextval(''' || destseq || ''') AS ' || const.pkey;
```
rafatower: past this point, we need a `destseq` always

rochoa:
Thanks for the heads up :-)

scripts-available/CDB_CartodbfyTable.sql
pramsey: You're throwing away existing primary key values here, on purpose?

test/CDB_CartodbfyTableTest.sql
```diff
@@ -225,6 +225,36 @@ SELECT CDB_CartodbfyTable('original');
...
+  the_geom geometry(Geometry,4326),
+  the_geom_webmercator geometry(Geometry,3857),
+  name text,
+  description text
+);
+INSERT INTO existing_cartodb_id (cartodb_id, description) VALUES
+       (1, 'a'),
+       (2, 'b'),
+       (3, 'c');
+SELECT CDB_CartodbfyTableCheck('existing_cartodb_id', 'Existing cartodb_id values are respected #138');
```
pramsey: To test that existing values are respected, use something like 100, 200, 300, not 1,2,3

scripts-available/CDB_CartodbfyTable.sql
rafatower: nope, big mistake on my side :(

thanks

test/CDB_CartodbfyTableTest.sql
rafatower: ok

scripts-available/CDB_CartodbfyTable.sql
pramsey: Phew, I almost didn't comment :) I assumed you'd only do that for some larger purpose I didn't understand :)

PR #151, opened 2015-09-15
First version of the cartodbfy requirements

rafatower:
@javisantana @pramsey  can you please review?

This is probably not complete and I know we don't meet some requirements yet, in particular "Cartodbfy shall deal with text columns for imports, regarding CartoDB columns" (which I think is important for importing CSV files without type guessing).

cc/ @iriberri

### DIFF
--- a/doc/cartodbfy-requirements.rst
+++ b/doc/cartodbfy-requirements.rst
@@ -1,0 +1,59 @@
+Introduction
+============
+
+This document aims at describing what cartodbfy is and what its formal requirements are, with the following goals in mind:
+
+- clarify what are the expectations of the "cartodbfycation process".
+- define an important part of what should be a stable, public API
+- allow for better testing, which should in turn...
+- ...ease modifications and increase quality of the code
+
+
+
+What is the cartodbfycation
+===========================
+
+The cartodbfycation is the process of converting an arbitrary postgres table into a valid CartoDB table, and register it in the system so that it can be used in the CartoDB editor and platform to generate maps and analysis.
+
+
+
+Valid CartoDB tables
+====================
+
+A valid CartoDB table shall meet the following conditions:
+
+- Have a ``cartodb_id`` column as primary key with a sequence
+- Have a ``the_geom`` column of type ``Geometry`` with SRID 4326
+- Have a ``the_geom_webmercator`` column of type ``Geometry`` with SRID 3857
+- The columns ``the_geom`` and ``the_geom_webmercator`` shall be in sync
+
+Additionally, a CartoDB table can contain other columns.
+
+
+
+High level requirements
+=======================
+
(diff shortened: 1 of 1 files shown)


javisantana:
FYI, there is a previous version, check the wiki: https://github.com/CartoDB/cartodb-postgresql/wiki/CartoDB-user-table

iriberri:
:clap: :clap:

doc/cartodbfy-requirements.rst
```diff
@@ -0,0 +1,59 @@
...
+- Exporting and re-importing the same table in CartoDB shall produce equivalent tables, with the same features associated to the same ``cartodb_id``'s.
+
+Note that there should be only one feature per row in the source table. If there's more than one, then which one is used for ``the_geom`` and ``the_geom_webmercator`` fields is not determined.
+
+
+
+Low-level requirements
+======================
+
+- If the original table contains a valid ``cartodb_id`` column, it shall be used
```
pramsey: "valid" is a magic word. more detail: integer, unique

doc/cartodbfy-requirements.rst
```diff
@@ -0,0 +1,59 @@
...
+- If the original table contains a valid ``the_geom`` or ``the_geom_webmercator`` it shall be used.
```
pramsey: in this case, "valid" = in the expected projection

doc/cartodbfy-requirements.rst
```diff
@@ -0,0 +1,59 @@
...
+The cartodbfycation is the process of converting an arbitrary postgres table into a valid CartoDB table, and register it in the system so that it can be used in the CartoDB editor and platform to generate maps and analysis.
+
+
+
+Valid CartoDB tables
+====================
+
+A valid CartoDB table shall meet the following conditions:
+
+- Have a ``cartodb_id`` column as primary key with a sequence
```
pramsey: integer requirement, sequence as default value

doc/cartodbfy-requirements.rst
rafatower: right
but text is also valid as long as unique, not null. currently not checking for non-null

iriberri:
I have applied Paul's and Rafa's feedback here. Merging @rafatower?

rafatower:
I'd add a deprecation note in https://github.com/CartoDB/cartodb-postgresql/wiki/CartoDB-user-table

other than that, ok. We can improve from here.

thanks for taking care of this long, overdue task.

iriberri:
Not null issue edited, wiki deprecated.

iriberri:
@rafatower is this okay now?

rafatower:
Yep, I think so.

BTW if you look into the "deprecated" doc, you'll find concerns about the multi-geometry thing we've talked about. I have my views on that (prefer not to comment here).

PR #152, opened 2015-09-15
141 fix table multiple geometry columns

rafatower:
Changes in this PR:
- A test for the failing scenario (maybe a naive one, but still better than none)
- Extracted `_cdb_geom_candidate_columns` query to a function.
- Added a `_cdb_has_usable_geom_record` type to be returned by `_CDB_Has_Usable_Geom`
- Pulled code that altered the table out of `_CDB_Has_Usable_Geom` and moved into the rewrite function.
- Modified how the table is rewritten so that the more generic type `GEOMETRY` is used instead of the specific one (POINT, MULTIPOINT, LINES etc...) from the original geom column. I have doubts about that and would like @javisantana to review them. I think formerly there was logic to cast MULTIPOINT to POINT but not sure about that or what's the expected behavior.
- Some small details: quoting of params, checks after EXECUTE, etc.

@pramsey can you please review?


scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -912,12 +928,25 @@ BEGIN
...
-      text_geom_column_srid boolean)
   INTO STRICT gc;
 
+  -- If geom is the wrong name, just rename it.
+  IF gc.has_geom AND gc.has_geom_name != const.geomcol THEN
+    sql := Format('ALTER TABLE %s DROP COLUMN IF EXISTS %I', reloid::text, const.geomcol);
+    PERFORM _CDB_SQL(sql,'_CDB_Rewrite_Table');
+    sql := Format('ALTER TABLE %s RENAME COLUMN %I TO %I', reloid::text, gc.has_geom_name, const.geomcol);
+    PERFORM _CDB_SQL(sql,'_CDB_Rewrite_Table');
+  END IF;
```
pramsey: So here (and below) you're OK with the dropping any existing non-geometry `the_geom` column if you've found an alternate geometry you can rename into place? I thought previously we were laundering potential collisions away by renaming the offending column `the_geom_1` or something.

scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -1070,15 +1099,11 @@ BEGIN
       )
       SELECT ', ST_Transform(' 
             || t.missing_srid_start || t.attname || t.missing_srid_end
-            || ',4326)::Geometry(' 
-            || t.geomtype 
-            || ',4326) AS '
+            || ',4326)::Geometry(GEOMETRY,4326) AS '
             || const.geomcol
```
pramsey: I take it the "enhancement" of retaining the original type restriction turned into a backwards compatibility problem :( There Shall Be No Changes Ever! :)

test/CDB_CartodbfyTableTest.sql
```diff
@@ -253,6 +253,18 @@ SELECT CDB_CartodbfyTableCheck('existing_cartodb_id', 'Existing cartodb_id value
...
+       the_geom geometry,
+       wkb_geometry geometry(MultiPoint,4326),
+       description varchar
+);
+INSERT INTO many_geometry_columns (the_geom, wkb_geometry) VALUES
+       ('0104000020E61000000100000001010000007108B023698052C03CEEA53A2E5D4440', '0104000020E61000000100000001010000007108B023698052C03CEEA53A2E5D4440'),
+       ('0104000020E6100000010000000101000000864C9E57618052C0994F0C7F3C5B4440', '0104000020E6100000010000000101000000864C9E57618052C0994F0C7F3C5B4440');
+SELECT CDB_CartodbfyTableCheck('many_geometry_columns', 'Table with both the_geom and wkb_geometry #141');
+SELECT * FROM many_geometry_columns;
+DROP TABLE many_geometry_columns;
```
pramsey: How about 
```
CREATE TABLE many_colliding_columns (
  the_geom varchar,
  the_geom_webmercator varchar,
  my_geom geometry,
  my_mercgeom geometry(Point, 3857),
  cartodb_id varchar,
  my_pk integer primary key
);

INSERT INTO many_colliding_columns VALUES (
 'foo', 'bar', 'SRID=4326;POINT(0 0)', 'SRID=3857;POINT(0 0)', 'nerf', 1
);
```

scripts-available/CDB_CartodbfyTable.sql
javisantana: tables in cartodb are not limited to a single geometry, don't confuse what cartodb wizards are manage with what the platform is able to provide

scripts-available/CDB_CartodbfyTable.sql
rafatower: where?

scripts-available/CDB_CartodbfyTable.sql
rafatower: then a table with the_geom and the_geom_webmercator with MultiPolygon restriction can be accepted as "perfect table in the perfect place" and be limited to a single geometry...

test/CDB_CartodbfyTableTest.sql
rafatower: okay, that's another test for the regression

scripts-available/CDB_CartodbfyTable.sql
javisantana: What would happen if we say "cartodb tables accept any kind of geometry", the user tries to insert a point and it fails? (does not sounds so perfect)

scripts-available/CDB_CartodbfyTable.sql
rafatower: Both drops were added in order to pass the new test. The renames were originally placed at the end of _CDB_Has_Usable_Geom

I double-checked the code and the prior renames are limited to dealing with cartodb_id column.

In practice I found this useful in cases where the source table has the_geom in a text type and some other such as wkb_geom with an actual geom that only needs renaming. Arguably we should rename instead of dropping.

scripts-available/CDB_CartodbfyTable.sql
pramsey: There's no real downside to having geometry unrestricted by type, since the only renderer hitting the tables in cartodb is one that we expect to be able to deal with multi-type inputs; so yeah, sorry for the noise, just being a big baby.

scripts-available/CDB_CartodbfyTable.sql
pramsey: Isn't the first step here running `DROP COLUMN` on any pre-existing `the_geom` column?

pramsey: If we aren't saving them now, I see no reason to preserve them going forward, so also ignore this comment :)

scripts-available/CDB_CartodbfyTable.sql
rafatower: Actually I kind of dropped that requirement in this changeset: 010dd13e

But turns out it never fully worked and also there's a problem with the check function calculating the expectations when there's a the_geom but not a the_geom_webmercator.

Issue #160, opened 2015-09-24
An upgrade from 0.7.3 to 0.10.1 fails


azamorano:
```
# ALTER EXTENSION cartodb UPDATE TO '0.10.1';
ERROR:  cannot change return type of existing function
HINT:  Use DROP FUNCTION cdb_usertables(text) first.
```

rafatower:
Done.

For the record:
- I updated the branch 3.11.X in cartodb. See https://github.com/CartoDB/cartodb/commit/56a83be5478190413cca8abf672cd3b03c841953
- I updated the cdb tag to point to 0.10.2. See https://github.com/CartoDB/cartodb-postgresql/releases

PR #161, opened 2015-09-24
Add DROP FUNCTION back to allow migrations #160

rafatower:
@rochoa please review

Just FYI: what changed between 0.7.4 and 0.8.0 was the return type of that function.

I'll test the upgrades by checking out that specific revision, installing and then going up to 0.10.2. I'll merge as soon as tests get passed (both clinker and those manual tests).

cc/ @azamorano

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 # cartodb/Makefile
 
 EXTENSION = cartodb
-EXTVERSION = 0.10.1
+EXTVERSION = 0.10.2
 
 SED = sed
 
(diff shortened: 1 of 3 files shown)


rafatower:
```
tests=# select cdb_version();
 cdb_version 
-------------
 0.7.3
(1 row)

tests=# alter extension cartodb update to '0.10.2';
ALTER EXTENSION
tests=# select cdb_version();
 cdb_version 
-------------
 0.10.2
(1 row)
```

PR #163, opened 2015-09-27
Write invalidation duration in postgresql log

luisbosque:
This is the equivalent in PL/pgSQL to the code proposed in https://github.com/CartoDB/cartodb/pull/5006

Instead of logging invalidation duration inside the python code, do it in the parent function. This way, the time van be logged no matter what type of invalidation is called after.

Using log raise level with plpy is the more appropiate since LOG level is much lower in log_min_messages (http://www.postgresql.org/docs/9.3/static/runtime-config-logging.html#GUC-LOG-MIN-MESSAGES) than in client_min_messages (http://www.postgresql.org/docs/9.3/static/runtime-config-logging.html#GUC-CLIENT-MIN-MESSAGES).
With a proper logging configuration the invalidation time should not bother in the client responses but it will appear in the postgresql log.

By default when a RAISE is executed, the context of the execution is shown in the log. Getting the value before the raise and setting it after does the trick.

This writes a line in the log like this:

Sep 27 17:20:44 cdb postgres[16384]: [53-1] 2015-09-27 17:20:44 GMT postgres cartodb_dev_user_889e2c90-9901-4c51-85b0-dcf93b02ace5_db psql LOG:  invalidation_duration: 0.001687

Related to https://github.com/CartoDB/cartodb/pull/5006

/cc @rochoa @javisantana

### DIFF
--- a/scripts-available/CDB_TableMetadata.sql
+++ b/scripts-available/CDB_TableMetadata.sql
@@ -64,6 +64,10 @@ DECLARE
   tabname TEXT;
   rec RECORD;
   found BOOL;
+  function_start timestamptz;
+  function_end timestamptz;
+  function_duration float;
+  log_error_verbosity text;
 BEGIN
 
   IF TG_OP = 'UPDATE' or TG_OP = 'INSERT' THEN
@@ -72,6 +76,8 @@ BEGIN
     tabname = OLD.tabname;
   END IF;
 
+  EXECUTE 'SELECT clock_timestamp()' INTO function_start;
+ 
   -- Notify table data update
   -- This needs a little bit more of research regarding security issues
   -- see https://github.com/CartoDB/cartodb/pull/241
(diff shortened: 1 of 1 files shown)


javisantana:
changing the log settings does not look right for me but not sure if there is another way cc @pramsey

scripts-available/CDB_TableMetadata.sql
```diff
@@ -72,6 +76,8 @@ BEGIN
     tabname = OLD.tabname;
   END IF;
 
+  EXECUTE 'SELECT clock_timestamp()' INTO function_start;
+ 
   -- Notify table data update
```
pramsey: `SELECT INTO function_start clock_timestamp()` doesn't work? (ie: no `EXECUTE`)

pramsey: Actually, `function_start := clock_timestamp();` should work.

pramsey:
It's a little worrisome, but parameters that are user-settable at run-time generally only affect one backend at a time, so apart from the ugliness it shouldn't matter. Do try to get rid of those `execute` calls though, I think you should be able to do it all directly.

scripts-available/CDB_TableMetadata.sql
```diff
@@ -105,7 +111,15 @@ BEGIN
...
   IF NOT found THEN RAISE WARNING 'Missing cdb_invalidate_varnish()'; END IF;
-
+  
+  EXECUTE 'SELECT clock_timestamp()' INTO function_end;
+  SELECT extract(epoch from (function_end - function_start)) INTO function_duration;
+  
+  EXECUTE 'SELECT setting FROM pg_settings where name=''log_error_verbosity''' INTO log_error_verbosity;
+  SET log_error_verbosity=TERSE;
+  RAISE LOG 'invalidation_duration: %', function_duration::text;
+  EXECUTE 'SET log_error_verbosity= ' || log_error_verbosity;
```
rafatower: I wouldn't touch log verbosity from within, but don't have a strong opinion against.

rafatower:
:+1: after a quick chat: I do not see a better option for logging without cluttering the log as the verbosity cannot be set by level as far as we could see in the docs.

javisantana:
\o/

rochoa:
:heart:

PR #164, opened 2015-09-27
New function to set the invalidate function name

luisbosque:
This is a probably useless proposal. I did it because before writing the invalidation duration log at https://github.com/CartoDB/cartodb-postgresql/pull/163 I was mistakenly thinking that every invalidation function cartodb has (telnet, http and service) had different names. However after writing it I realized that the function name was always the same, and just the content changed.

Just proposing here in case someone thinks it may be useful instead of directly sending it to the recycle bin. If not, please close de PR.

In case somebody things it's useful, just note it's WIP

/cc @rochoa


rafatower:
I'm closing this PR, it will be still searchable in github in case we need it.

PR #167, opened 2015-09-29
move alaska, hawaii, and puerto rico into frame with lower 48 states all in an albers

andy-esch:
This function projects U.S. states and Puerto Rico into a frame that's easy to view for election maps, etc.

Sample usage:

``` sql
update states_table
set the_geom = CDB_AlbersUSA(the_geom,state_identifier)
```

Produces a map like this:

screen shot 2015-09-28 at 21 56 55

I'm looking for feedback on:
1. Choosing appropriate projections
2. Placement and scale of the states
3. Whether more state identifiers should be listed for the second argument

cc @makella @andrewxhill

### DIFF
--- a/scripts-available/CDB_Albers.sql
+++ b/scripts-available/CDB_Albers.sql
@@ -9,30 +9,23 @@
 --  output: geometries of states in albers projections of the states
 -- 
 -- Projections:
--- 		- 
+--		- Lower 48 states: http://spatialreference.org/ref/sr-org/7965/
+-- 		- Alaska: http://spatialreference.org/ref/epsg/3338/
 --		- Puerto Rico: http://www.spatialreference.org/ref/epsg/nad83-puerto-rico-virgin-is/
 --		- Hawaii: http://epsg.io/102007
 
 
-CREATE OR REPLACE FUNCTION CDB_AlbersUSA (g geometry, state text, srid numeric DEFAULT 3857) RETURNS geometry as $$ 
+CREATE OR REPLACE FUNCTION CDB_Albers50 (g geometry, state text) RETURNS geometry as $$
 DECLARE
 	reply geometry;
-	new_srid INT;
-	alaska text[];
-	hawaii text[];
-	puertorico text[];
+	srid INT;
+	alaska text[] = '{"Alaska","AK","02"}'::text[];
+	hawaii text[] = '{"Hawaii","HI","15"}'::text[];
+	puertorico text[] = '{"Puerto Rico","PR","72"}'::text[];
 BEGIN
 	
-	alaska = '{"Alaska","AK","02"}'::text[];
-	hawaii = '{"Hawaii","HI","15"}'::text[];
-	puertorico = '{"Puerto Rico","PR","72"}'::text[];
-	
-	-- decide on SRID to use
-	IF srid = 3857 THEN 
-		new_srid = ST_SRID(g);
-	ELSE
-		new_srid = srid;
-	END IF;
+	-- convert to wgs84
+	IF ST_SRID(g) != 4326 THEN g = ST_Transform(g,4326); END IF;
 
 	EXECUTE 'SELECT
 	    ST_SetSRID(
@@ -82,10 +75,10 @@ BEGIN
 				ELSE
 					ST_Transform($1,42303)
   				END
-  				, $6
+  				, 3857
   			)'
 	INTO reply
-	USING g, state, alaska, hawaii, puertorico, new_srid;
+	USING g, state, alaska, hawaii, puertorico;
 
 	RETURN reply;
 
--- a/scripts-available/CDB_Albers.sql
+++ b/scripts-available/CDB_Albers.sql
@@ -8,13 +8,13 @@
 --  output: geometries of states in albers projections of the states
 -- 
 -- Projections:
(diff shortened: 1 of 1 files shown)


andrewxhill:
Does it work line your example, overwriting thegeom, it does it only work
if you alias the result as thegeomweb?

ohasselblad:
I can easily change it to be agnostic to the input geometry, which might be more natural: the srid you put in is the srid you get out.

updating the_geom_webmercator works if it's transformed:

```sql
update states_table
set the_geom_webmercator = ST_Transform(CDB_AlbersUSA(the_geom,state_identifier),3857)
```

But that doesn't trigger `the_geom` column to be recalculated.

andrewxhill:
Yep, got it got, I missed the srid parameter. Looks great!

On Tue, Sep 29, 2015 at 12:12 PM, Andy Eschbacher <notifications@github.com>
wrote:

> I can easily change it to be agnostic to the input geometry, which might
> be more natural: the srid you put in is the srid you get out.
>
> updating the_geom_webmercator works if it's transformed:
>
> update states_tableset the_geom_webmercator = ST_Transform(CDB_AlbersUSA(the_geom,state_identifier),3857)
>
> But that doesn't trigger the_geom column to be recalculated.
>
> —
> <https://github.com/CartoDB/cartodb-postgresql/pull/167#issuecomment-144107376>
> .
>



-- 

andrewxhill <http://twitter.com/andrewxhill>
http://cartodb.com

ohasselblad:
@rochoa @javisantana, what do you think of this? Also, how should I test it?

javisantana:
Why this needs to be a function in the extension and not a table in common data?

ohasselblad:
Yeah. Where should I save this function for later reference, then?

andrewxhill:
Because any user that wants to transform their own data to match will need
to perform the same transform exactly. (Imagine any point dataset)

javisantana:
I think we should do something more generic here, we know a lot of countries have this problem, so why now having a table with the codes <-> transformations (and projection), like

```
CDB_Country('spain', the_geom, admin_identifier)
CDB_Country('usa', the_geom, admin_identifier)
````

and the a table like

```
create table cdb_countries (
  name text,
  region text,
  offset, 
  projection_srid
)
```

maybe sounds overkill but adding so concrete functionality to a core extension seems weird (not against it tho)

**Comment by ohasselblad** [comment]
Ok, I'll:

+ make a list of candidate countries (or continents) that could use this type of transformation
+ look into the 'standard' way to alter the projection
+ fit it into the format proposed above

**Comment by makella** [comment]
@ohasselblad there was a comment from Cindy B about the Census maps that she was a part of making... apparently AK (the state itself) prefers to have AK above the contiguous US. I know you and I had discussed the idea of having a couple of different layouts for templates including what USGS and Census do.

On a design note: she also said that I should make sure to shadow in Canada, Mexico in the background.

Example of one her maps:

screen shot 2015-10-19 at 8 34 34 am

**Comment by andrewxhill** [comment]
i see this working when you have control of the non-map layout too, but for
more generic purposes i feel like we still need to move alaska below. space
packing + familiarity

On Mon, Oct 19, 2015 at 8:39 AM, Mamata Akella <notifications@github.com>
wrote:

> @ohasselblad <https://github.com/ohasselblad> there was a comment from
> Cindy B about the Census maps that she was a part of making... apparently
> AK (the state itself) prefers to have AK above the contiguous US. I know
> you and I had discussed the idea of having a couple of different layouts
> for templates including what USGS and Census do.
>
> On a design note: she also said that I should make sure to shadow in
> Canada, Mexico in the background.
>
> Example of one her maps:
>
> [image: screen shot 2015-10-19 at 8 34 34 am]
> <https://cloud.githubusercontent.com/assets/1566273/10577844/41e17a0a-763c-11e5-90da-fbc0766331af.png>
>
> —
> <https://github.com/CartoDB/cartodb-postgresql/pull/167#issuecomment-149202101>
> .
>



-- 

andrewxhill <http://twitter.com/andrewxhill>
http://cartodb.com

**Comment by ohasselblad** [comment]
And I did make Alaska larger than it is in most other projections like this!

**Comment by nerik** [comment]
Hey there, I'm happy to unearth this :)
I've been using this, very useful and thank you 
https://team.cartodb.com/u/nerikcarto/viz/36e410e6-0574-11e6-9b33-0e31c9be1b51/embed_map

One thing that is noticed when mapping the Zika outbreak: american territories are missing (American Samoas and US Virgin Islands). [They are important, no?](https://www.youtube.com/watch?v=CesHr99ezWE) 

This would be awesome if merged - even with only the USA, but a with generic name.

**Comment by nerik** [comment]
A quick way to add separator lines :
```
    -- puerto rico
    SELECT ST_Transform(ST_GeomFromText('LINESTRING(13 1.5,16 3)', 4236), 3857) as the_geom_webmercator
    union all
    -- hawaii
    SELECT ST_Transform(ST_GeomFromText('LINESTRING(-12 7.8,-5 1.6)', 4236), 3857) as the_geom_webmercator
    union all
    -- alaska
    SELECT ST_Transform(ST_GeomFromText('LINESTRING(-21 10,-9 -1)', 4236), 3857) as the_geom_webmercator
```


screen shot 2016-05-12 at 19 19 14

Algunenano:
Stale

PR #168, opened 2015-09-29
148 cartodbfy checks for null cartodb id

### LINKED ISSUE #148: cartodbfy shall discard cartodb_id candidates that contain nulls
Currently it checks for uniqueness but not for nullity.

gfiorav:
Fixes #148

CDB_CartodbfyTable() now checks if the found `cartodb_id` column contains null ids in addition to previously checked uniqueness. This should avoid crashes when importing any files containing a `cartodb_id` column containing null values.

Tests that check this functionality have also been added.

CR please @rafatower :smiley: 

cc @rochoa


test/CDB_CartodbfyTableTest.sql
```diff
@@ -281,6 +281,44 @@ INSERT INTO many_colliding_columns VALUES (
...
+-- table with null cartodb_id
+CREATE TABLE test (
+  cartodb_id integer
+);
+INSERT INTO test VALUES
+  (1), 
+  (2), 
+  (NULL), 
+  (3);
+SELECT CDB_CartodbfyTable('test');
```
rafatower: Use `CDB_CartodbfyTableCheck` instead, that provides extra checks on the resulting table after cartodbfy'ing

rafatower:
:+1: but let's wait for @rochoa so we can decide how to roll out this fix

PR #169, opened 2015-10-05
Changes CDB_DateToNumber() to accept timestamps with time zones 2202

gfiorav:
Fixes [#2202](https://github.com/CartoDB/cartodb/issues/2202)

@rafatower 

The trick is that this way if a timestamp without a time zone is submitted it will still work.

CR and thoughts please! :smile:


rafatower:
That's a public function, part of the API and we shouldn't simply change it.

Actually you are creating a different function. In a test DB (creating it in `public` schema):
```sql
CREATE OR REPLACE FUNCTION CDB_DateToNumber(input timestamp)               
RETURNS double precision AS $$
DECLARE output double precision;
BEGIN
    BEGIN
        SELECT extract (EPOCH FROM input) INTO output;
    EXCEPTION WHEN OTHERS THEN
        RETURN NULL;
    END;
RETURN output;
END;
$$
LANGUAGE 'plpgsql' STABLE STRICT;
CREATE FUNCTION

CREATE OR REPLACE FUNCTION CDB_DateToNumber(input timestamp with time zone)
RETURNS double precision AS $$
DECLARE output double precision;
BEGIN
    BEGIN
        SELECT extract (EPOCH FROM input) INTO output;
    EXCEPTION WHEN OTHERS THEN
        RETURN NULL;
    END;
RETURN output;
END;
$$
LANGUAGE 'plpgsql' STABLE STRICT;
CREATE FUNCTION


\df public.cdb_datetonumber               
                                     List of functions
 Schema |       Name       | Result data type |        Argument data types        |  Type  
--------+------------------+------------------+-----------------------------------+--------
 public | cdb_datetonumber | double precision | input timestamp with time zone    | normal
 public | cdb_datetonumber | double precision | input timestamp without time zone | normal
(2 rows)
```

In this particular case, we can take advantage of [function overloading](http://www.postgresql.org/docs/9.4/static/sql-createfunction.html#SQL-CREATEFUNCTION-OVERLOADING).

Please:
- split it in two functions
- add tests for this change, as naive as they could be.

cc/ @rochoa

gfiorav:
@rafatower 

Done!

test/CDB_DateToNumberTest.sql
```diff
@@ -0,0 +1,2 @@
+SELECT * FROM CDB_DateToNumber('1999-01-08 00:00:00+00'::timestamp);
```
rafatower: semantically, the `+00` is not natural for a timestamp, so if we want to check specifically for non-timezone I'd remove it

test/CDB_DateToNumberTest.sql
```diff
@@ -0,0 +1,2 @@
...
+SELECT * FROM CDB_DateToNumber('1999-01-08 00:00:00+00'::timestamp with time zone);
```
rafatower: and here just the opposite, it is nice to check a non-utc input, right?

gfiorav:
@rafatower my approach was:

```sql
tests=# create table test (date timestamp with time zone);
CREATE TABLE
tests=# insert into test values ('1999-01-08');
INSERT 0 1
tests=# select * from test;
          date          
------------------------
 1999-01-08 00:00:00+00
(1 row)

```

(I used `1999-01-08` because it was an example in the [PostgreSQL documentation](http://www.postgresql.org/docs/9.1/static/datatype-datetime.html))

So my assumption is that `1999-01-08 00:00:00+00` is the de facto string representation of a `TIMESTAMP WITH TIME ZONE` type.

In the first line, I cast it to a normal timestamp to test that the old code works (`CDB_DateToNumber(timestamp)`).
In the second line, I cast to a timestamp with time zone to test that the new code works.

Does it make sense?

rafatower:
For me the first case definitely doesn't make sense even if it works, because as you said you're using the string representation of a `timestamp with timezone` and then cast down to a `timestamp` without timezone.

Let your code speak out its intent as clearly as possible.

gfiorav:
@rafatower 
You're right: the more intuitive the better.

See if you like the new tests better.

rafatower:
Fine by me, but as in the other case, let's wait till tomorrow to see how we'll deal with the release cycle. Thanks!

gfiorav:
Great, thank you!

PR #173, opened 2015-10-21
Better unique identifiers

### LINKED ISSUE #172: [CartoDBfy] Long name table provokes "relation already exists" error
First of all, I think this is something we are already aware of from different issues but I don't know if this is being treated specifically somewhere as I'm not able to find the issues now. Sorry if duplicated, we found it again during a regression, so here it is in case it's helpful. 

**STR**:
1. Upload [this file](https://gist.githubusercontent.com/iriberri/0c2a66f146fb6f574874/raw/29f81d4add74b280067af790b565c974596ba65e/ksdjovheufhvosdsijdsdsfiuhdsfhuidfshiudfsiuhdsfdfddsfsfsduhvofehfivedsvidv.csv) by URL (or download it respecting the name)

**Result**:
````
2015-10-15 14:59:18 UTC: Proceeding to register 2015-10-15 14:59:18 UTC: Before renaming from importer_4dd2d1c0734d11e5ac170e98b61680bf to ksdjovheufhvosdsijdsdsfiuhdsfhuidfshiudfsiuhdsfdfddsfsfsduhvofehfivedsvidv 
2015-10-15 14:59:18 UTC: Before moving schema 'ksdjovheufhvosdsijdsdsfiuhdsfhuidfshiudfsiuhdsfdfddsfsfsduhvofe' from cdb_importer to iriberri 
2015-10-15 14:59:18 UTC: Before persisting metadata 'ksdjovheufhvosdsijdsdsfiuhdsfhuidfshiudfsiuhdsfdfddsfsfsduhvofe' data_import_id: 9d7369eb-713f-4730-b07a-263a4235c7cf 
2015-10-15 14:59:20 UTC: Exception: PG::Error: ERROR: CDB(_CDB_Rewrite_Table:42P07:relation "ksdjovheufhvosdsijdsdsfiuhdsfhuidfshiudfsiuhdsfdfddsfsfsduhvofe" already exists): CREATE SEQUENCE iriberri.ksdjovheufhvosdsijdsdsfiuhdsfhuidfshiudfsiuhdsfdfddsfsfsduhvofe_cartodb_id_seq C
````

### PR DESCRIPTION
**Pull request by gfiorav**
Fixes #172

There are now three new functions: `_CDB_Unique_Identifier`, `_CDB_Unique_Column_Identifier` and `CDB_Octet_Trim` that are used to find unique identifiers taking into account length restrictions (63 octets as in v9.4). They're also UTF8 safe (accounts that different character may have different octet lengths).

This deprecates two non-UTF8 non-length aware methods for finding identifiers:
- `_CDB_Unique_Relation_Name`
- `_CDB_Unique_Column_Name`

I’m porting the old code that uses these two functions as well. 

Still working on some stuff so this is not ready yet.

To do:
- [x] CDB_Unique_Identifier
- [x] CDB_Unique_Column_Identifier
- [x] CDB_Octet_Truncate
- [x] Move to CDB_Helper.sql
- [x] Port code using `_CDB_Unique_Relation_Name`
- [x] Port code using `_CDB_Unique_Column_Name `
- [x] Test for CDB_Helper.sql
- [x] CR
- [x] Regression tests

### CODE REVIEW

**Comment by gfiorav** [comment]
I think this is ready for a CR, @rafatower

File: scripts-available/CDB_CartodbfyTable.sql [hunk 1]
```diff
@@ -427,8 +428,9 @@ END;
...
 
 
--- Find a unique relation name in the given schema, starting from the 
--- template given. If the template is already unique, just return it; 
+-- DEPRECATED: Use cartodb.CDB_Unique_Identifier since it's UTF8 Safe and length
+-- aware. Find a unique relation name in the given schema, starting from the
+-- template given. If the template is already unique, just return it;
 -- otherwise, append an increasing integer until you find a unique variant.
 CREATE OR REPLACE FUNCTION _CDB_Unique_Relation_Name(schemaname TEXT, relationname TEXT)
 RETURNS TEXT
```
rafatower: I'd raise an error on deprecated functions.

scripts-available/CDB_Helper.sql
```diff
@@ -0,0 +1,163 @@
+-- UTF8 safe and lenght aware. Find a unique identifier with a given prefix
+-- and/or suffix and withing a schema.
+CREATE OR REPLACE FUNCTION cartodb.CDB_Unique_Identifier(prefix TEXT, relname TEXT, suffix TEXT, schema TEXT DEFAULT NULL)
```
rafatower: these should be "private" functions, with the `_CDB` prefix

test/CDB_HelperTest.sql
```diff
@@ -0,0 +1,64 @@
...
+DROP TABLE piñaácidpiñaácidpiñaácidpiñaácidpiñaácidpi_suffix;
+
+-- Test CDB_Trim_Octets simple case
+SELECT * FROM cartodb.CDB_Octet_Trim('piraña', 1);
+
+-- Test CDB_Octet_Trim UTF8 case
+SELECT * FROM cartodb.CDB_Octet_Trim('piraña', 2);
+
+-- Test CDB_Octet_Trim UTF8 case
+SELECT * FROM cartodb.CDB_Octet_Trim('piraña', 3);
```
rafatower: there are no tests for `CDB_Unique_Column_Identifier`

gfiorav:
If you're happy with CR rectified version, I can merge this and start staging regression tests @rafatower 

Thanks!

scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -439,6 +441,8 @@ DECLARE
   newrelname TEXT;
 BEGIN
 
+  RAISE EXCEPTION '_CDB_Unique_Relation_Name is DEPRECATED. Use _CDB_Unique_Identifier(prefix TEXT, relname TEXT, suffix TEXT, schema TEXT DEFAULT NULL)';
+
   i := 0;
```
rafatower: minor detail: the old code is to be deleted

scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -479,6 +484,8 @@ DECLARE
   newcolname TEXT;
 BEGIN
 
+  RAISE EXCEPTION '_CDB_Unique_Column_Name is DEPRECATED. Use _CDB_Unique_Column_Identifier(prefix TEXT, relname TEXT, suffix TEXT, reloid REGCLASS DEFAULT NULL)';
+
   i := 0;
   newcolname := columnname;
```
rafatower: same here

scripts-available/CDB_Helper.sql
```diff
@@ -0,0 +1,163 @@
+-- UTF8 safe and lenght aware. Find a unique identifier with a given prefix
+-- and/or suffix and withing a schema.
+CREATE OR REPLACE FUNCTION cartodb._CDB_Unique_Identifier(prefix TEXT, relname TEXT, suffix TEXT, schema TEXT DEFAULT NULL)
```
rafatower: It is customary to add an explanation of the arguments' semantic. Important in some cases, for instance: meaning of the default `schema DEFAULT NULL`.

scripts-available/CDB_Helper.sql
```diff
@@ -0,0 +1,163 @@
...
+    i := i + 1;
+  END LOOP;
+
+  PERFORM _CDB_Error('looping too far', '_CDB_Unique_Identifier');
+END;
+$$ LANGUAGE 'plpgsql';
+
+-- UTF8 safe and lenght aware. Find a unique identifier for a column with a given prefix
+-- and/or suffix and withing a realtion. If no reloid is give, all relations are examined
+CREATE OR REPLACE FUNCTION cartodb._CDB_Unique_Column_Identifier(prefix TEXT, relname TEXT, suffix TEXT, reloid REGCLASS DEFAULT NULL)
```
rafatower: Does examining all relations make sense?

scripts-available/CDB_Helper.sql
```diff
@@ -0,0 +1,163 @@
...
+CREATE OR REPLACE FUNCTION cartodb._CDB_Unique_Identifier(prefix TEXT, relname TEXT, suffix TEXT, schema TEXT DEFAULT NULL)
+RETURNS TEXT
+AS $$
+DECLARE
+  rec RECORD;
+  usedspace INTEGER;
+  ident TEXT;
+  i INTEGER;
+  origident TEXT;
+  maxlen INTEGER;
```
rafatower: Better defining a constant to avoid programming mistakes: http://www.postgresql.org/docs/9.4/static/plpgsql-declarations.html

scripts-available/CDB_Helper.sql
```diff
@@ -0,0 +1,163 @@
...
+BEGIN
+  maxlen := 63;
+
+  usedspace := 3;
```
rafatower: what's this `3`?

rafatower: ok, seems to be the max length of the numeric suffix. Worth a comment?

scripts-available/CDB_Helper.sql
gfiorav: This will only happen if no reloid is specified. If you think it's useless I'll just remove the `DEFAULT NULL`.

scripts-available/CDB_Helper.sql
gfiorav: The suffix `_xx` that can be added at the end if the identifier is already taken.

scripts-available/CDB_Helper.sql
```diff
@@ -0,0 +1,163 @@
...
+    IF reloid IS NOT NULL THEN
+      SELECT a.attname
+      INTO rec
+      FROM pg_class c
+      JOIN pg_attribute a ON a.attrelid = c.oid
+      WHERE NOT a.attisdropped
+      AND a.attnum > 0
+      AND c.oid = reloid
+      AND a.attname = ident;
+    ELSE
```
rafatower: I'd completely remove the DEFAULT NULL and the ELSE block

scripts-available/CDB_Helper.sql
```diff
@@ -0,0 +1,163 @@
...
+    i := i + 1;
+  END LOOP;
+
+  PERFORM _CDB_Error('looping too far', '_CDB_Unique_Column_Identifier');
+END;
+$$ LANGUAGE 'plpgsql';
+
+-- Trims the end of a given string by the given number of octets taking care
+-- not to leave characters in half. UTF8 safe.
+CREATE OR REPLACE FUNCTION cartodb._CDB_Octet_Trim(tostrip TEXT, octets INTEGER)
```
rafatower: How about changing `octets` by `octets_to_tream`? Would make it a bit more readable IMHO

scripts-available/CDB_Helper.sql
```diff
@@ -0,0 +1,163 @@
...
+  origident TEXT;
+  maxlen INTEGER;
+BEGIN
+  maxlen := 63;
+
+  usedspace := 3;
+  usedspace := usedspace + coalesce(octet_length(prefix), 0);
+  usedspace := usedspace + coalesce(octet_length(suffix), 0);
+
+  relname := _CDB_Octet_Trim(relname, usedspace + octet_length(relname) - maxlen);
```
rafatower: called like `_CDB_Unique_Column_Identifier('foo', 'tbl', 'bar');` then:
```
usedspace + octet_length(relname) - maxlen = 9 + 3 - 63 = -51
```

is that ok?

scripts-available/CDB_Helper.sql
gfiorav: I'd expect `Octets_to_trim` to return the amount of octets that have to trimmed. 

Maybe `Octet_Substring`?

scripts-available/CDB_Helper.sql
```diff
@@ -0,0 +1,163 @@
...
+  trimmed TEXT;
+BEGIN
+  charlen := bit_length('a');
+  tostriplen := char_length(tostrip);
+  expected := tostriplen * charlen;
+  examined := bit_length(tostrip);
+
+  IF expected = examined OR octets = 0 THEN
+    RETURN SUBSTRING(tostrip from 1 for (tostriplen - octets));
+  ELSIF octets < 0 THEN
```
rafatower: IMHO the case of `octets` less than OR equal to zero should be checked first, and documented as part of the function's interface.

scripts-available/CDB_Helper.sql
rafatower: I saw below it is ok. But better a minor correction and a bit of function documentation via small comments (see below)

scripts-available/CDB_Helper.sql
rafatower: this function returns a `TEXT` so I don't know how you expect to return that

rafatower:
I made some comments, the bulk of it is ok and more robust than what we have before. I'd strive for polishing it a little further if you don't mind before merging. Regarding testing in staging, you don't need merging into master and is better not to.

Merging is good when:
- have several things to release, and
- you're pretty confident on not breaking anything with them.

(we now the process is slow, but it is the best we have).

scripts-available/CDB_Helper.sql
gfiorav: I was just pointing out that `Octets_to_trim` _syntacticly_ seems to return a numerical quantity, or even maybe perform a conversion. If I saw a variable called `octets_to_trim`, it would make perfect sense to me it contained the amount of octets to be trimmed, maybe it's just me :smile: 

Would `Trim_Octets` make sense for you? It might clear the purpose of the method more that way

gfiorav:
Ok, thanks for the CR!

Regarding merging into master, I was just following the docs https://github.com/CartoDB/cartodb-management/wiki/Deploy-cartodb-postgresql-extension#develop. Should we modify them? They encourage to merge to master before testing in Staging.

gfiorav:
@rafatower Went for `CDB_Octet_Truncate` :smile: 

I think it's much better now. See if you agree.

scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -961,19 +918,18 @@ BEGIN
...
-  destseq := _CDB_Unique_Relation_Name(destschema, destseq);
+  destseq := cartodb._CDB_Unique_Identifier(NULL, relname, '_' || const.pkey || '_seq', destschema);
   destseq := Format('%I.%I', destschema, destseq);
   PERFORM _CDB_SQL(Format('CREATE SEQUENCE %s', destseq), '_CDB_Rewrite_Table');
 
   -- Salt a temporary table name if we are re-writing in place
   -- Note copyname is already escaped and safe to use as identifier
   IF destschema = relschema THEN
-    copyname := Format('%I.%I', destschema, Format('%s_%s', destname, salt));
+    copyname := Format('%I.%I', destschema, cartodb._CDB_Unique_Identifier(NULL, destname, '_' || salt), destschema);
```
rafatower: I would remove the salting thing, it was done just and precisely to get unique names, apparently (sorry I did not notice before).

gfiorav: No prob, good catch!

scripts-available/CDB_Helper.sql
```diff
@@ -0,0 +1,154 @@
...
+  END LOOP;
+
+  PERFORM _CDB_Error('looping too far', '_CDB_Unique_Identifier');
+END;
+$$ LANGUAGE 'plpgsql';
+
+
+-- UTF8 safe and lenght aware. Find a unique identifier for a column with a given prefix
+-- and/or suffix and withing a realtion.
+CREATE OR REPLACE FUNCTION cartodb._CDB_Unique_Column_Identifier(prefix TEXT, relname TEXT, suffix TEXT, reloid REGCLASS)
```
rafatower: another late catch: relname and reloid are redundant and thus calling for problems.

I'd stick to reloid whenever possible, as in the deprecated function.

scripts-available/CDB_Helper.sql
```diff
@@ -0,0 +1,154 @@
...
+  relname := _CDB_Octet_Truncate(relname, maxlen - usedspace);
+
+  IF relname = '' THEN
+    PERFORM _CDB_Error('prefixes are to long to generate a valid identifier', '_CDB_Unique_Column_Identifier');
+  END IF;
+
+  ident := coalesce(prefix, '') || relname || coalesce(suffix, '');
+
+  i := 0;
+  origident := ident;
```
rafatower: you can directly use
```
origident := coalesce(prefix, '') || relname || coalesce(suffix, '');
```
and remove the assignemt above

scripts-available/CDB_Helper.sql
```diff
@@ -0,0 +1,154 @@
...
+  origident TEXT;
+
+  i INTEGER;
+BEGIN
+  -- Accounts for the _XX incremental suffix in case the identifier is taken
+  usedspace := 3;
+  usedspace := usedspace + coalesce(octet_length(prefix), 0);
+  usedspace := usedspace + coalesce(octet_length(suffix), 0);
+
+  relname := _CDB_Octet_Truncate(relname, maxlen - usedspace);
```
rafatower: I wouldn't modify the original relname received as a param. Anyway, as pointed above I think it is better to remove it as a param and keep it as a temp variable.

scripts-available/CDB_Helper.sql
gfiorav: Ok, it's a typo. `relname` is `colname`. I need a base column name AND the relation within which guarantee the identifiers uniqueness.

scripts-available/CDB_Helper.sql
gfiorav: But I would need some way of preserving the original truncated version so I can use it as a base to append `_XX`. I'm I missing something?

scripts-available/CDB_Helper.sql
```diff
@@ -0,0 +1,154 @@
...
+    RETURN string;
+  END IF;
+
+  strlen := char_length(string);
+
+  expected := char_length(string);
+  examined := octet_length(string);
+
+  IF expected = examined THEN
+    RETURN SUBSTRING(string from 1 for max_octets);
```
rafatower: I'd always use lowercase for standard function names (actually for all of them).

can't you use `left` instead?

http://www.postgresql.org/docs/9.4/static/functions-string.html

scripts-available/CDB_Helper.sql
rafatower: you're right

PR #175, opened 2015-10-30
Removes extra d in delimiter 155

### LINKED ISSUE #155: Correct wrong format strings
```
contrib_regression=# select format('CREATE SEQUENCE %s START WITH %d', 'hola', 3+2);
ERROR:  unrecognized conversion type specifier "d"
contrib_regression=# select format('CREATE SEQUENCE %s START WITH %i', 'hola', 3+2);
ERROR:  unrecognized conversion type specifier "i"
contrib_regression=# select format('CREATE SEQUENCE %s START WITH %s', 'hola', 3+2);
```

This affects current code: https://github.com/CartoDB/cartodb-postgresql/blob/master/scripts-available/CDB_CartodbfyTable.sql#L168

gfiorav:
Fixes #155

To Do:

- [x] Fix bug
- [x] Write tests
- [x] CR (@rafatower)

### DIFF
--- a/scripts-available/CDB_CartodbfyTable.sql
+++ b/scripts-available/CDB_CartodbfyTable.sql
@@ -152,7 +152,7 @@ BEGIN
       -- Copy existing values to new field
       -- NOTE: using ALTER is a workaround to a PostgreSQL bug and is also known to be faster for tables with many rows
       -- See http://www.postgresql.org/message-id/20140530143150.GA11051@localhost
-      sql := Format('ALTER TABLE %s ALTER cartodb_id TYPE int USING %I', reloid::text, new_name);
+      sql := Format('ALTER TABLE %s ALTER cartodb_id TYPE int USING %I::integer', reloid::text, new_name);
       RAISE DEBUG 'Running %', sql;
       EXECUTE sql;
 
--- a/scripts-available/CDB_CartodbfyTable.sql
+++ b/scripts-available/CDB_CartodbfyTable.sql
@@ -157,7 +157,7 @@ BEGIN
       EXECUTE sql;
 
       -- Find max value
-      sql := Format('SELECT max(cartodb_id) FROM %s', reloid::text);
+      sql := Format('SELECT coalesce(max(cartodb_id), 0) as max FROM %s', reloid::text);
       RAISE DEBUG 'Running %', sql;
       EXECUTE sql INTO rec;
 
@@ -166,7 +166,7 @@ BEGIN
         AS seq INTO rec2;
 
       -- Reset sequence name
-      sql := Format('ALTER SEQUENCE %s RESTART WITH %', rec2.seq::text, rec.max + 1);
+      sql := Format('ALTER SEQUENCE %s RESTART WITH %s', rec2.seq::text, rec.max + 1);
       RAISE DEBUG 'Running %', sql;
       EXECUTE sql;
 
--- a/scripts-available/CDB_CartodbfyTable.sql
+++ b/scripts-available/CDB_CartodbfyTable.sql
@@ -166,7 +166,7 @@ BEGIN
         AS seq INTO rec2;
 
       -- Reset sequence name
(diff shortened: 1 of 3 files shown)


scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -166,7 +166,7 @@ BEGIN
         AS seq INTO rec2;
 
       -- Reset sequence name
-      sql := Format('ALTER SEQUENCE %s RESTART WITH %d', rec2.seq::text, rec.max + 1);
+      sql := Format('ALTER SEQUENCE %s RESTART WITH %', rec2.seq::text, rec.max + 1);
```
rafatower: I don't think that code is correct
```
# select format('test %s %', 'foo', 42);
ERROR:  unterminated format specifier
```
Have you checked that the test actually fails without the fix? I don't think you're exercising that line with the test.

Have you checked there are no other %X wrong format specifiers?

gfiorav:
@rafatower there was a problem where the `rec`'s `max` field wasn't set. It is now working.

test/CDB_CartodbfyTableTest_expect
```diff
@@ -102,5 +102,12 @@ Table with non unique and null cartodb_id #148 cartodbfied fine
 3|
 4|2
 DROP TABLE
+CREATE TABLE
+
+DROP TABLE
+CREATE TABLE
+
+_cartodb_id0
+DROP TABLE
```
rafatower: I'd set the verbosity to NOTICE in the tests above to make sure the message "Could not initialize cartodb_id with existing values" is not raised during the tests.

gfiorav:
@rafatower 

There is no `notice` level. http://www.postgresql.org/docs/9.4/static/app-psql.html

I'm setting it to `terse`.

rafatower:
@gfiorav there are verbosity levels and logging levels http://www.postgresql.org/docs/9.0/static/runtime-config-logging.html

set the verbosity level:
```
\set VERBOSITY [terse|default|verbose]
```
set the logging level:
```
SET client_min_messages TO [debug|...];
```

gfiorav:
Ah! I see.

Thanks for the info. I'll change that tomorrow

gfiorav:
Ok, @rafatower, find the code refactored to your suggestions.

Much appreciated!

test/CDB_CartodbfyTableTest_expect
```diff
@@ -102,5 +102,21 @@ Table with non unique and null cartodb_id #148 cartodbfied fine
...
+NOTICE:  Column cartodb_id already exists
+NOTICE:  Existing cartodb_id field does not have an associated sequence, renaming
+NOTICE:  Trying to recover data from _cartodb_id0 column
+
+DROP TABLE
+CREATE TABLE
+NOTICE:  Column cartodb_id already exists
+NOTICE:  Existing cartodb_id field is of invalid type text (need int2, int4 or int8), renaming
+NOTICE:  Trying to recover data from _cartodb_id0 column
+NOTICE:  Could not initialize cartodb_id with existing values: result of USING clause for column "cartodb_id" cannot be cast automatically to type integer (42804)
```
rafatower: this is exactly the message that should not be raised :((

rafatower: https://github.com/CartoDB/cartodb-postgresql/pull/175/files#diff-767fb5c7568f72bdf3da71fbe0d21ca4R180

gfiorav: Oh shoot! I apologise for the oversight. Kind of overloaded atm. I'll be back with it soon.

gfiorav:
@rafatower It seems that the original code was missing an explicit `::integer` cast for the `ALTER TABLE %s ALTER cartodb_id TYPE int USING %I::integer` expression. Now it should be OK.

rafatower:
:+1:

PR #180, opened 2015-11-24
174 cartodbfy

rafatower:
Creating this PR on behalf of @jgoizueta 
- It allows dots in column names to be cartodbfied
- It fixes a corner case when adding sequences if the table already had created a sequence (hitting when moving users in some cases)


rafatower:
:+1:

PR #182, opened 2015-12-03
CDB_DotDensity

stuartlynn:
Adding a function CDB_DotDensity(the_geom, no_points) that will randomly fill a polygon with the desired number of points. An example of the output created from census tract data is [here](http://bit.ly/1OH45BK) 

screen shot 2015-12-03 at 3 16 56 pm

Works pretty well but that example took about 3 minutes to run. The main problem is that the function generates points randomly within the bounding box of the polygon then rejects them if they aren't  contained within the polygon. This requires a lot of ST_Contains calls which are expensive and the misses will get high for long thin diagonal polygons.

Looking for suggestions about how to do this smarter. Some ideas 
- Call ST_Simplify on the polygon before starting the process of generating the points. Not sure if ST_Contains scales with the polygon complexity
- Subdivide the bounding box in to a grid of n x n boxes. Check if a box lies entirely outside the polygon , if so then discard it. If the box lies entirely within the polygon randomly populate it with points proportional to the fraction of the parent polygon's area that the box takes up. If the box is partly inside the polygon populate it randomly with points proportional to the intersection area of this box with the poly who pass a ST_Contains test. 
- Something smart with the quad key index? 

@andrewxhill  @pramsey would love your thoughts.


stuartlynn:
Leaving this PR for discussion DONT MERGE just yet

stuartlynn:
Actually this might be a better approach: 

https://gist.github.com/stuartlynn/2ac9cbc0a0a4e3fba239

pramsey:
Using St_Contains() is fine, as long as you use it in a context that allows the prepared geometry code to get leverage, and that would be in a SQL statement formed like a join, something like:
```sql
SELECT pts.pt FROM 
(select geom) as geom, 
(select st_makepoint( random(), random()) as pt from generate_series(1,n)) as pts
WHERE st_contains(geom, pt)
```
And then the loop would be about doing more random points for until the quota was full.

Would still have problems w/ long skinny diagonals though.

ohasselblad:
Closing this as it's going to be in crankshaft -> https://github.com/CartoDB/crankshaft/pull/148

PR #185, opened 2015-12-18
Overviews

rafatower:
@pramsey We mostly concur with your comments regarding aggregation of other attributes (categorical values, modes and average).

Let's focus on the points reduction for the moment (`_CDB_GridCluster_Reduce_Strategy`). I also understand the "spare loner" over-representation problem you mentioned. Is there any other thing algorithm-wise or performance-wise that you think we could improve?

Thanks!

cc/ @jgoizueta


rafatower:
I forgot to mention that spatial clustering plus sampling could work well IMHO...

scripts-available/CDB_Overviews.sql
```diff
@@ -0,0 +1,220 @@
+CREATE OR REPLACE FUNCTION _CDB_Feature_Density(reloid REGCLASS, nz integer)
```
pramsey: So help me out here. This is exploratory, looking for the zoom level at which the feature density on a "typical" tile exceeds some threshold?

scripts-available/CDB_Overviews.sql
```diff
@@ -0,0 +1,220 @@
...
+  SELECT CDB_XYZ_Resolution(-8) INTO c;
+
+  -- We first compute a set of *seed* tiles, of the minimum Z level, z0, such that
+  -- they cover the extent of the table and we have at least n of them in each
+  -- linear dimension (i.e. at least n*n tiles cover the extent).
+  -- We compute the number of features in these tiles, and recursively in
+  -- subtiles up to level z0 + nz. Then we compute the maximum of the feature
+  -- density (per tile area in webmercator squared meters) for all the
+  -- considered tiles.
+  EXECUTE Format('
```
pramsey: Is it possible there's a premature optimization going on here? You know the extent of the table, the number of features therein, the average and distribution of feature areas... you can probably make a good guess at which zoom level the feature density is going to exceed some tolerance, without doing a big pile of recursive queries.

scripts-available/CDB_Overviews.sql
```diff
@@ -0,0 +1,220 @@
...
+      WHERE t.e > %2$s AND t.z < (base.z + %3$s)
+    )
+    SELECT MAX(e/ST_Area(CDB_XYZ_Extent(x,y,z))) FROM t where e > 0;
+  ', reloid::text, min_features, nz, n, c)
+  INTO fd;
+  RETURN fd;
+  END
+$$ LANGUAGE PLPGSQL STABLE;
+
+CREATE OR REPLACE FUNCTION _CDB_Dummy_Ref_Z_Strategy(reloid REGCLASS)
```
pramsey: So the "base z level" is the level from which overviews will be built, yes?

pramsey:
It looks pretty good to me. I like the pluggability of it all, nice work.

scripts-available/CDB_Overviews.sql
javisantana: Another possibility is to use postgis index stats, what a pity they can't be easily interpreted without using C ( cc @pramsey). 

This query will likely run on tables with million of records, doing counts and full scans are going to make it really slow. We could go in the same way postgis does, to have cdb_density and cdb_estimated_density as postgis has st_estimatedextent

In general too many magic numbers (like n = 4, 256)

scripts-available/CDB_Overviews.sql
rafatower: right

scripts-available/CDB_Overviews.sql
```diff
@@ -0,0 +1,220 @@
...
+    IF is_overview THEN
+      SELECT substring(ref::text FROM 1 FOR length(ref::text)-length(suffix)) INTO base;
+    ELSE
+      base := ref;
+    END IF;
+    RETURN Format('%s_ov%s', base::text, overview_z);
+  END
+$$ LANGUAGE PLPGSQL IMMUTABLE;
+
+CREATE OR REPLACE FUNCTION _CDB_Dummy_Reduce_Strategy(reloid REGCLASS, ref_z INTEGER, overview_z INTEGER)
```
javisantana: Could we change the name to sampled strategy?

scripts-available/CDB_ZoomFromScale.sql
```diff
@@ -1,10 +1,9 @@
 CREATE OR REPLACE FUNCTION cartodb.CDB_ZoomFromScale(scaleDenominator numeric) RETURNS int AS $$
```
rochoa: This is coming from: https://github.com/CartoDB/CartoDB-basemaps/blob/7ca728b1a7b5d0d5c999c9eac1521e95eb5a4446/global_functions.sql#L77-L107

If we plan to change this, what about using https://github.com/mapnik/mapnik/wiki/ScaleAndPpi as reference to calculate the proper values?

See also: https://github.com/openstreetmap/mapnik-stylesheets/blob/236f8d0ab17b792300679cedf9afa5b0feab301b/zoom-to-scale.txt

jgoizueta: I think the current behaviour should be preserved for this function: Mapnik provides a `scale-denominator` value which is unique for all the tiles of a fixed zoom level, although the true scale will vary from tile to tile (with latitude). That is good for us, because we can unambiguosly derive the zoom level from that single scale value, so we can know which zoom level is being rendered with a given query that uses `!scale-denominator!`.

If we were to change this, having Mapnik provide the actual scale of each tile, we should either use a different token name, keeping the current behaviour of `scale-denominator` or add another token `zoom-level` which could be used in the cases where `CDB_ZoomFromScale` is now being used.

jgoizueta:
@pramsey @rafatower  please CR this. cc @javisantana 

I think we've got something usable here, we have `CDB_CreateOverviews(table)` to create overviews for (points) table and `CDB_DropOverviews(table)` to remove them.

Then, there's `CDB_Overviews(table)` to obtain information for using the overviews in rendering as shown in these examples:

* http://bl.ocks.org/jgoizueta/a054e9f345ac28ada9e5
* http://bl.ocks.org/jgoizueta/881fedb9b83df2b75f60

javisantana:
I don't think CDB_Overviews is a good idea, cache wise we will have a lot of problems. Why not to have methods to know the tables and some other stuff instead of hiding the stuff behind a function?

jgoizueta:
My first intention was to maintain some overviews metadata tabla, but I decided to delay decisions about that by having the CDB_Overviews() interface to that metadata. 

Considering caching opportunities (which I hadn't) it makes sense to eliminate the interface function and go for having the metadata table to be used directly by clients.

javisantana:
I think makes sense to have more granularity methods there like:

````
# select * CDB_Overviews('table')

zoom |   table
------+-----------
   10 | tablename

# select * CDB_OverviewAggType('table')

CDB_OverviewAggType
---------------------------------
cluster
```

so we can play with that form the tiler.

We can talk about that tomorrow

scripts-available/CDB_Overviews.sql
```diff
@@ -0,0 +1,517 @@
...
+    EXECUTE Format('ALTER TABLE IF EXISTS %s OWNER TO %I;', overview_table::text, table_owner);
+
+    -- preserve the table privileges
+    UPDATE pg_class c_to
+      SET  relacl = c_from.relacl
+      FROM  pg_class c_from
+      WHERE c_from.oid  = dataset
+      AND   c_to.oid    = overview_table;
+
+    PERFORM _CDB_Add_Indexes(overview_table);
```
rafatower: minor comment: this doesn't look like the responsibility of register

PR #187, opened 2015-12-21
Overviews rtorre touches

rafatower:
@jgoizueta I've written this code while taking a closer look to yours. Feel free to comment and/or merge into `overviews` branch.

Cheers!

### DIFF
--- a/scripts-available/CDB_Overviews.sql
+++ b/scripts-available/CDB_Overviews.sql
@@ -1,3 +1,40 @@
+-- Calculate the estimated extent of a cartodbfy'ed table.
+-- Scope: private.
+-- Parameters
+--   reloid: oid of the input table.
+-- Return value A box2d extent in 3857.
+CREATE OR REPLACE FUNCTION _cdb_estimated_extent(reloid REGCLASS)
+RETURNS box2d
+AS $$
+  DECLARE
+    ext box2d;
+    ext_query text;
+    table_id record;
+  BEGIN
+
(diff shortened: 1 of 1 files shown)


rafatower:
Some comments:
- it seems to be asymptotic both in exec times and returned results (which kind of makes sense)
- for low zoom levels it tends to underestimate the max zoom returned. From the [docs](http://www.postgis.org/docs/ST_Estimated_Extent.html):

> resulting extent will be about 95% of the real one

scripts-available/CDB_Overviews.sql
```diff
@@ -1,3 +1,46 @@
...
+      table_id.schema_name, table_id.table_name, 'the_geom_webmercator'
+    );
+
+    BEGIN
+      EXECUTE ext_query INTO ext;
+      EXCEPTION
+        -- This is the typical ERROR: stats for "mytable" do not exist
+        WHEN internal_error THEN
+          -- Get stats and execute again
+          EXECUTE format('ANALYZE %1$I', reloid);
```
pnorman: `ANALYZE %1$I (the_geom_webmercator)` might be better here, as it only analyzes the column needed for this function. On the other hand, if one column is missing statistics, they probably all are, and analyzing everything could be useful.

pnorman:
> for low zoom levels it tends to underestimate the max zoom returned

The 2-d postgis case is a bit more complicated than sampling numbers, but for large tables, postgres basically uses sampling of random pages, which means that it's likely to miss some maxima.

rafatower:
merged as per our conversation

PR #188, opened 2015-12-22
Convert some simple functions from plpgsql to sql

pnorman:
SQL is a faster language to call, and these are very simple functions.

I checked with psql, and changing the language doesn't change the function signature

### DIFF
--- a/scripts-available/CDB_XYZ.sql
+++ b/scripts-available/CDB_XYZ.sql
@@ -5,24 +5,9 @@
 CREATE OR REPLACE FUNCTION CDB_XYZ_Resolution(z INTEGER)
 RETURNS FLOAT8
 AS $$
-DECLARE
-  earth_circumference FLOAT8;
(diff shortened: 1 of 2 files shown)


scripts-available/CDB_XYZ.sql
```diff
@@ -5,24 +5,9 @@
...
-  tile_size := 256;
-
-  full_resolution := earth_circumference/tile_size;
-
-  RETURN full_resolution / (power(2,z));
-
-END
-$$ LANGUAGE 'plpgsql' IMMUTABLE STRICT;
+  -- circumference divided by 256 is z0 resolution, then divide by 2^z
+  SELECT 40075017 / 256 / power(2, z);
```
jgoizueta: Should be 40075017.0 to be taken as floating point and match the former results. That's the reason for the tests failing.

I will be merging this into the overviews branch so it is included in the next release.

PR #189, opened 2015-12-23
Fix test errors on Travis

pnorman:
This adjusts the output of one test to what Travis is producing, and rounds the output of another test.

The second test looked like it could have differed depending on floating point details, so hopefully round fixes that.


rafatower:
:+1:

PR #194, opened 2016-02-03
[WIP] Add CDB_ForeignTable management functions

zenitraM:
So this is the first draft :sparkles: of how using `postgres_fdw` to access tables on other database (a CartoDB database) would work.

All of the magic comes from a json set in the CDB_Conf parameter `fdws`. It contains something like:

```
{  
   "foo":{  
      "server":{  
         "extensions":"postgis",
         "dbname":"cartodb_dev_user_7dfdbeff-97bc-4258-9fff-7bb55ced269e_db",
         "host":"localhost",
         "port":"5432"
      },
      "users":{  
         "public":{  
            "user":"fdw_foo_user",
            "password":"fdw_foo_pw123"
         }
      }
   }
}
```

With this, calling the function `_CDB_Create_FDW('foo');` will, given the enough permissions:
- Create the `FOREIGN SERVER`, called `foo`
- Set its parameters as according to the `server` properties
- Create a `USER MAPPING` for everybody in `users` (the "public" in this example is the origin user. Using "public" is a special user which is used as a default one by all users, not sure if we want to connect as anybody else) and set its properties.
- Create a `foo` schema.
- Grant permissions to use it to the organization role
- Bring the `cartodb.cdb_tablemetadata` remote table as `foo.cdb_tablemetadata`

This function is idempotent and can be ran as many times as desired - I've tried to verify for pre-existing things as much as possible (i.e check if the schema exists instead of running `CREATE SCHEMA IF NOT EXISTS`) to make it run always without any warnings for pre-existing things.

There's also the wrapper function `CDB_Add_Remote_Table('foo', 'provinces');`. This will setup the `foo` fdw and, afterwards, add the `provinces` table as the foreign table `foo.provinces`. (This one should not exist previously).
And it will also grant SELECT access to it to publicuser.

`CDB_Add_Remote_Table` has right now a security definer, so that any ordinary user can bring any remote table of a FDW an admin has configured for him. :sparkles:

This is a first PoC, probably the function names could change (Setup_FDW instead of Create_FDW to indicate the idempotency?). Some other pending concerns I can think of:
- Security: who do we want to give read access to the table? who should be able to call AddRemoteTable?. Need to review the permissions this needs / grants.
- Tests and doc :sparkles:, have to figure out a way to test/CI this.
- Compatibility: This does not use FDW table inheritance (per se) so it should work with 9.3. FDW extension pushdown (supported from 9.6 onward, or our backported 9.5) is only enabled when the "extension: postgres" is set on the config, so without it I believe it should work on 9.3, but that's something I'll figure out when I get to the tests.
- Oddities with using a foreign cdb_tablemetadata. postgres_fdw does some sort of remote OID to local OID translation, so that if you import cdb_tablemetadata only but have no tables it fails to cross-reference them like this:
```
cartodb_dev_user_6f72fe3d-6aee-413e-9cb7-0b4049c86dc2_db=# SELECT * FROM "do".cdb_tablemetadata;
ERROR:  relation "do.ign_spanish_adm3_municipalities" does not exist
CONTEXT:  column "tabname" of foreign table "cdb_tablemetadata"
```
But I'm not sure if querying the table for a specific table will work, it might do.

@javisantana @rafatower @rochoa

### DIFF
--- a/scripts-available/CDB_ForeignTable.sql
+++ b/scripts-available/CDB_ForeignTable.sql
@@ -4,7 +4,7 @@
 -- All the FDW settings are read from the `cdb_conf.fdws` entry json file.
 ---------------------------
 
-CREATE OR REPLACE FUNCTION cartodb._CDB_Create_FDW(name text, config json)
+CREATE OR REPLACE FUNCTION cartodb._CDB_Setup_FDW(fdw_name text, config json)
 RETURNS void
 AS $$
 DECLARE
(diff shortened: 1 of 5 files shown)


javisantana:
> CDB_Add_Remote_Table has right now a security definer, so that any ordinary user can bring any remote table of a FDW an admin has configured for him.

Not sure we want to allow this, I'd say to keep this function only for postgres user

scripts-available/CDB_ForeignTable.sql
```diff
@@ -0,0 +1,122 @@
...
+
+CREATE OR REPLACE FUNCTION cartodb._CDB_Create_FDW(name text, config json)
+RETURNS void
+AS $$
+DECLARE
+  row record;
+  option record;
+  org_role text;
+BEGIN
+  IF NOT EXISTS ( SELECT * FROM pg_extension WHERE extname = 'postgres_fdw')
```
rafatower: minor thing: I personally find more readable to have IF...END in the same line

scripts-available/CDB_ForeignTable.sql
```diff
@@ -0,0 +1,122 @@
+---------------------------
+-- FDW MANAGEMENT FUNCTIONS
+--
+-- All the FDW settings are read from the `cdb_conf.fdws` entry json file.
+---------------------------
+
+CREATE OR REPLACE FUNCTION cartodb._CDB_Create_FDW(name text, config json)
```
rafatower: minor thing: there are a few var/params called `name` in this file. I'd rename this one to be called `srvname` or similar for clarity's sake

scripts-available/CDB_ForeignTable.sql
```diff
@@ -0,0 +1,122 @@
...
+        ELSE
+          EXECUTE FORMAT('ALTER USER MAPPING FOR %I SERVER %I OPTIONS (SET %I %L)', row.key, name, option.key, option.value);
+        END IF;
+      END LOOP;
+    END LOOP;
+
+    -- Create schema if it does not exist.
+    IF NOT EXISTS ( SELECT * from pg_namespace WHERE nspname=name)
+      THEN
+      EXECUTE FORMAT ('CREATE SCHEMA %I', name);
```
rafatower: (the schema sharing the name with the server and the config is an implementation detail)

scripts-available/CDB_ForeignTable.sql
```diff
@@ -0,0 +1,122 @@
...
+  FOR row IN SELECT p.key, p.value from lateral json_each(cartodb.CDB_Conf_GetConf('fdws')) p
+    LOOP
+      EXECUTE 'SELECT cartodb._CDB_Create_FDW($1, $2)' USING row.key, row.value;
+    END LOOP;
+  END
+  $$
+  LANGUAGE PLPGSQL;
+
+
+CREATE OR REPLACE FUNCTION cartodb._CDB_Create_FDW(name text)
```
rafatower: minor thing: replace `name` by `server_name`?

rafatower:
We shall focus on figuring out how to access the foreign cdb_tablemetadata (or equivalent).

scripts-available/CDB_ForeignTable.sql
zenitraM: Right now there are three things which share the same name:
- the config entry name
- the schema name
- the `FOREIGN SERVER` name

I wanted to keep them the same as a convention, but you're right about `name` being too generic. What about `fdw_name`?

zenitraM:
I've been fighting today with reading from cdb_tablemetadata, as the column type is `regclass` it made FDW queries a pain until I found the workaround, querying the column casted to text.

Also I've figured out how to use introspection to discover the remote table/schema name.

Right now there's a [CDB_Get_Foreign_Updated_at](https://github.com/CartoDB/cartodb-postgresql/commit/b957635e78e123adbd6e31b99a3b3bc01665952c#diff-d1587e6ad59f6d6aad72af7f0d7ffeabR112) which will get the remote table updated_at, but we will also probably need the remote schema, remote name and remote database to have the proper cache key c-c-c-combo breaker combination.

rafatower:
I'm working under the assumption that sql and tiler would need `{dbname, schema, table_name, updated_at}`.

Without having tested `CDB_Get_Foreign_Updated_at` nor `CDB_QueryTablesText` yet (but almost done with the painful setup) I'd say I'd need to make some changes in this branch.

rafatower:
If I execute as `postgres`:
```
# SELECT cartodb.CDB_Add_Remote_Table('foo', 'us_census_tiger2013_state');
ERROR:  user mapping not found for "postgres"
```

and if I execute as `cdb`, as there's no longer SECURITY DEFINER:
```
SELECT cartodb.CDB_Add_Remote_Table('foo', 'us_census_tiger2013_state');
ERROR:  permission denied for relation cdb_conf
```
I'll patch my local branch with security definer but will need to find a soln. for this.

zenitraM:
I've added some specs, and figured out (by the Travis run) that on`CDB_Add_Remote_Table` _won't work_ with 9.3 due to it not having `IMPORT FOREIGN SCHEMA` :disappointed: 

I've done a workaround to skip it by manually creating the foreign table on the specs on versions <9.5. CDB_Setup_FDW works and is always tested, though.

rafatower:
:+1: 

@javisantana what's your view on the compat issues with 9.3?

zenitraM:
Should we make `CDB_Query_Tables_Updated_At` quote the names that need quoting in SQL? see https://github.com/CartoDB/Windshaft-cartodb/pull/374#issuecomment-182368709

rafatower:
IMHO it is best to quote/escape on input rather than on output. But that's debatable. I'm ok with doing it, but we need to be careful with double-quoting:
```
my_extension_test=# CREATE TABLE "test-with-dash" ();
CREATE TABLE

my_extension_test=# \dt "test-with-dash"
             List of relations
 Schema |      Name      | Type  |  Owner   
--------+----------------+-------+----------
 public | test-with-dash | table | postgres
(1 row)


my_extension_test=# SELECT 'test-with-dash'::regclass::text;
       text       
------------------
 "test-with-dash"

my_extension_test=# SELECT format('%I', 'public.test-with-dash'::regclass::text);                                                                                                                                                             
        format        
----------------------
 """test-with-dash"""
```

zenitraM:
Yes, that's because casting to `regclass` will cause it to be quoted, not the case of the new function:

```
cartodb_dev_user_6f72fe3d-6aee-413e-9cb7-0b4049c86dc2_db=# select format('%I', table_name) as table_name, format('%I', schema_name) as schema_name from cdb_querytablesupdatedat('select * from "fo-bar".lol');
 table_name | schema_name
------------+-------------
 lol        | "fo-bar"
(1 row)

cartodb_dev_user_6f72fe3d-6aee-413e-9cb7-0b4049c86dc2_db=# select table_name, schema_name from cdb_querytablesupdatedat('select * from "fo-bar".lol');
 table_name | schema_name
------------+-------------
 lol        | fo-bar
(1 row)
```

`QueryTablesText` returns it escaped due to its `regclass` cast, so I'd prefer to be consistent with it

rafatower:
fair enough

zenitraM:
no, I was wrong, QueryTables returns it quoted due to explicitly calling quote_ident:

https://github.com/CartoDB/cartodb-postgresql/blob/master/scripts-available/CDB_QueryTables.sql#L48

(Although I guess the same could be done with format(%I instead of %s + quote_ident)

rafatower:
My points are:
- ok with quoting it on output for consistency, and
- let's be careful with double-quoting.

the point I dropped:
- better to quote in the inputs rather than outputs (security guys common wisdom)

zenitraM:
Filed https://github.com/CartoDB/cartodb-postgresql/issues/198 for some other issue we can live with for now but I think after the quoting everything matches nicely.

rafatower:
After what we talked: I'm gonna pick changes at https://github.com/CartoDB/cartodb-postgresql/commit/e870cf96e090fd55c3c82156c262a7a77ccfa473 and create another PR from there.

then pick the windshaft branch https://github.com/CartoDB/Windshaft-cartodb/pull/374 and setup its tests in such a way they install `cartodb` extension functions in the `cartodb` schema and add it to the test user `search_path`.

rafatower:
Continuing here: https://github.com/CartoDB/cartodb-postgresql/pull/199

PR #196, opened 2016-02-08
Fdw query tables updated at

rafatower:
Sample usage:
```sql
SELECT * FROM CDB_QueryTablesUpdatedAt($quotesql$
  SELECT local.cartodb_id, local.the_geom, local.the_geom_webmercator, remote.name
  FROM test_sp_join_fdw as local, foo.us_census_tiger2013_state as remote
  WHERE st_contains(remote.the_geom, local.the_geom);
$quotesql$);
```

```
                         db_name                          | schema_name |        table_name         |          updated_at           
----------------------------------------------------------+-------------+---------------------------+-------------------------------
 cartodb_dev_user_36c4a45a-eb92-4af4-a8ff-1065ecfd041f_db | foo         | us_census_tiger2013_state | 2016-02-05 15:39:18.997476+00
 cartodb_dev_user_95dc74f3-84b8-4738-a16e-92bab6a9a9af_db | public      | test_sp_join_fdw          | 2016-02-05 15:37:19.744318+00
(2 rows)
```

Tests are pending but I'd appreciate some early feedback

cc/ @javisantana @zenitraM


zenitraM:
:+1:, although the APIs also support obtaining the latest updated_at date from the tables given a list of tables.
This is used to cache the affected tables and not having to parse again the query, but still fetching the latest updated_at date.

Maybe it makes sense to expose something like `CDB_UpdatedAt(tables []text)` so it can be used both standalone and combined with `CDB_QueryTablesText` output, and then make `CDB_QueryTablesUpdatedAt` combine both

cc @rochoa

rafatower:
I agree. I'll work on it and add tests

On Tue, Feb 9, 2016 at 10:44 AM, Alejandro Martínez <
notifications@github.com> wrote:

> [image: :+1:], although the APIs also support obtaining the latest
> updated_at date from the tables given a list of tables.
> This is used to cache the affected tables and not having to parse again
> the query, but still fetching the latest updated_at date.
>
> Maybe it makes sense to expose something like CDB_UpdatedAt(tables []text)
> so it can be used both standalone and combined with CDB_QueryTablesText
> output, and then make CDB_QueryTablesUpdatedAt combine both
>
> cc @rochoa <https://github.com/rochoa>
>
> —
> <https://github.com/CartoDB/cartodb-postgresql/pull/196#issuecomment-181785384>
> .
>



-- 
*Rafa de la Torre*
DataServices Engineer www.cartodb.com

rafatower:
Changes since last review:
- I renamed `CDB_QueryTablesUpdatedAt` to `CDB_QueryTables_Updated_At` (I really hate CamelCase when postgres is case-insensitive most of the time). I also renamed the returned param `db_name` to `dbname`. Pretty silly changes but affect integration with client code.
- Moved everything to `CDB_ForeignTable.sql`. Why's that? because those functions are fdw-aware (though they work if there's no foreign table) and they are coupled to our fdw standard assumption: _(schema_name, table_name) map to the remote ones with the same name_
- Small fixes to make tests and extension installation work
- Some tests (I abused yours a little).

@zenitraM can you please review again? If you're ok then I'll merge into `fdw` and prepare the release.

zenitraM:
LGTM :+1:

PR #197, opened 2016-02-09
Use standard formula for CDB_ZoomFromScale

pnorman:
postgis-vt-util comes with a standard formula for zoom from scaleDenominator, and this is well tested. This also fixes the function throwing an error on NULL input.

There is a slight behavior change, as some of the old thresholds were not correct, and the function errored on NULL input.

```
 scaledenominator | new_zoom | old_zoom
------------------+----------+----------
                0 |          |       23
       1000000000 |          |        0
        500000000 |        0 |        1
             1000 |       19 |       19
```

calling the function with scaleDenominator = 0 or > 600000000 indicates a bogus value, which generally happens when mapnik runs a query with LIMIT 0 to identify column names. The former has infinite zoom, and the latter has negative zoom, neither of which makes sense.

### DIFF
--- a/scripts-available/CDB_ZoomFromScale.sql
+++ b/scripts-available/CDB_ZoomFromScale.sql
@@ -1,30 +1,12 @@
-CREATE OR REPLACE FUNCTION cartodb.CDB_ZoomFromScale(scaleDenominator numeric) RETURNS int AS $$
-BEGIN
+CREATE OR REPLACE FUNCTION cartodb.CDB_ZoomFromScale(scaleDenominator numeric)
+RETURNS int
+LANGUAGE SQL
+IMMUTABLE
+AS $$
+SELECT
   CASE
-    WHEN scaleDenominator > 500000000 THEN RETURN 0;
-    WHEN scaleDenominator <= 500000000 AND scaleDenominator > 200000000 THEN RETURN 1;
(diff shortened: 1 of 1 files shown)


jgoizueta:
We'll have to check all the places where we substitute scaledenominator by 0 (for various purposes) to see if they are affected by the new NULL value of zoom:

* https://github.com/CartoDB/Windshaft-cartodb/blob/56095926e0d5f7723d8f342275ad36ac748a5c51/lib/cartodb/api/overviews_metadata_api.js#L18
* https://github.com/CartoDB/Windshaft/blob/2cd43d4b2ca2b3487ede8423f698b282c6e53af4/lib/windshaft/backends/attributes.js#L81
* https://github.com/CartoDB/Windshaft-cartodb/blob/overviews-work/lib/cartodb/api/query_tables_api.js#L92
* https://github.com/CartoDB/Windshaft/blob/2cd43d4b2ca2b3487ede8423f698b282c6e53af4/lib/windshaft/renderers/torque/factory.js#L245

jgoizueta:
With the current `CDB_ZoomFromScale` implementation the maximum zoom is 23. If we use the formula there's no a priori limit:

```
# select _test_ZoomFromScale(1e-1000);
 _test_zoomfromscale
---------------------
                3351
(1 row)
```

This could be convenient, e.g. for overviews queries (see #258)  but we should check we're not expecting bounded zoom values anywhere this function is used.

jgoizueta:
Close via #259

PR #200, opened 2016-02-11
Add CDB_TableMetadata_Text view as a proxy to access FDW tablemetadata

zenitraM:
`CDB_Get_Foreign_Updated_At` was assuming the `tabname::text` filter would be passed to the foreign backend, however the query planner decides sometimes that that filter is not worth pushing down and tries to fetch the whole table.

When doing so, it fails if there is a remote table that is not locally due to the regclass casting oddities:
```cartodb_dev_user_6f72fe3d-6aee-413e-9cb7-0b4049c86dc2_db=# explain analyze SELECT updated_at FROM "do".cdb_tablemetadata WHERE tabname::text='"do".ign_spanish_adm3_municipalities'::text LIMIT 1;
ERROR:  relation "do.ne_10m_admin_1_states_provinces_lines_shp" does not exist
CONTEXT:  column "tabname" of foreign table "cdb_tablemetadata"```

This introduces a new view `CDB_TableMetadata_Text` that has the contents of CDB_TableMetadata but casted to FQTN.

cc @rafatower

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ $(EXTENSION)--$(EXTVERSION).sql: $(CDBSCRIPTS) cartodb_version.sql Makefile
 	echo '\echo Use "CREATE EXTENSION $(EXTENSION)" to load this file. \quit' > $@
 	cat $(CDBSCRIPTS) | \
     $(SED) -e 's/public\./cartodb./g' \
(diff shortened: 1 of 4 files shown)


rafatower:
retest this, please

PR #202, opened 2016-03-01
Add support to detect string cartodb_id columns

iriberri:
Going to try adding more tests now.

### DIFF
--- a/scripts-available/CDB_CartodbfyTable.sql
+++ b/scripts-available/CDB_CartodbfyTable.sql
@@ -480,7 +480,6 @@ DECLARE
   i INTEGER;
   sql TEXT;
   useable_key BOOLEAN = false;
-  has_usable_primary_key BOOLEAN = false;
 BEGIN
 
   RAISE DEBUG 'CDB(_CDB_Has_Usable_Primary_ID): %', 'entered function';
--- a/scripts-available/CDB_CartodbfyTable.sql
+++ b/scripts-available/CDB_CartodbfyTable.sql
@@ -466,19 +466,33 @@ END;
 $$ LANGUAGE 'plpgsql';
 
 
+
+DO $$
+BEGIN
+  IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = '_cdb_has_usable_primary_id_record') THEN
+    CREATE TYPE _cdb_has_usable_primary_id_record
+      AS (has_usable_primary_key boolean,
+        text_key_column boolean);
+    END IF;
+END$$;
+
 -- Find out if the table already has a usable primary key
 -- If the table has both a usable key and usable geometry
 -- we can no-op on the table copy and just ensure that the 
 -- indexes and triggers are in place
-CREATE OR REPLACE FUNCTION _CDB_Has_Usable_Primary_ID(reloid REGCLASS)
-  RETURNS BOOLEAN
+CREATE OR REPLACE FUNCTION cartodb._CDB_Has_Usable_Primary_ID(reloid REGCLASS)
+  RETURNS _cdb_has_usable_primary_id_record
 AS $$
 DECLARE
   rec RECORD;
   const RECORD;
+  idc RECORD;
   i INTEGER;
   sql TEXT;
   useable_key BOOLEAN = false;
+  has_usable_primary_key BOOLEAN = false;
+  -- In case cartodb_id is a text column
+  text_key_column BOOLEAN = false;
 BEGIN
 
   RAISE DEBUG 'CDB(_CDB_Has_Usable_Primary_ID): %', 'entered function';
@@ -506,7 +520,8 @@ BEGIN
       -- And it's a unique primary key! Done!
       IF (rec.indisprimary OR rec.indisunique) AND rec.attnotnull THEN
         RAISE DEBUG 'CDB(_CDB_Has_Usable_Primary_ID): %', Format('found good ''%s''', const.pkey);
-        RETURN true;
+        SELECT true as has_usable_primary_key, text_key_column INTO idc;
+        RETURN idc;
 
       -- Check and see if the column values are unique and not null,
       -- if they are, we can use this column...
@@ -520,9 +535,10 @@ BEGIN
           RAISE DEBUG 'CDB(_CDB_Has_Usable_Primary_ID): Found text column %', rec.atttypid;
 
           BEGIN
-            sql := Format('ALTER TABLE %s ALTER cartodb_id TYPE int USING %I::integer', reloid::text, rec.attname);
+            sql := Format('SELECT %I::integer FROM %s', rec.attname, reloid::text);
             RAISE DEBUG 'CDB(_CDB_Has_Usable_Primary_ID): Running %', sql;
             EXECUTE sql;
+            text_key_column := true
             EXCEPTION
             WHEN invalid_text_representation THEN
               RAISE DEBUG 'CDB(_CDB_Has_Usable_Primary_ID): Column % of type text is not a valid integer column', rec.attname;
@@ -571,7 +587,8 @@ BEGIN
         
         END IF;
         
-        return useable_key;
+        SELECT useable_key as has_usable_primary_key, text_key_column INTO idc;
+        RETURN idc;
 
       END IF;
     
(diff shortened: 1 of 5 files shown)


scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -499,40 +499,57 @@ BEGIN
...
-            PERFORM _CDB_Error(sql, '_CDB_Has_Usable_Primary_ID');          
-        END;
-  
+
+        -- If the column is varchar, try to cast it to integer
+        IF rec.atttypid IN (1043) THEN
+          RAISE DEBUG 'CDB(_CDB_Has_Usable_Primary_ID): Found text column %', rec.atttypid;
+
+          BEGIN
+            sql := Format('ALTER TABLE %s ALTER cartodb_id TYPE int USING %I::integer', reloid::text, rec.attname);
```
rafatower: better noting this is traversing the table and therefore breaking the paradigm of the "rewrite" function: just do one pass after figuring it out how to transform into the intended target table.

scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -499,40 +499,57 @@ BEGIN
...
+            RAISE DEBUG 'Running %', sql;
+            EXECUTE sql;
+            EXCEPTION
+            WHEN others THEN
```
rafatower: better use specific exceptions when possible, as in rails

iriberri:
Related disabled tests: https://github.com/CartoDB/cartodb-postgresql/blob/54973c01f2a6561a728653f68b5e9b330048da12/test/CDB_CartodbfyTableTest.sql#L167-L179

rafatower:
you can blame me :S and re-enable those tests

iriberri:
Not blaming! text was not supported anymore so... :)

I reenabled and added an extra one for text column in an empty table, I'm still playing with it locally.

scripts-available/CDB_CartodbfyTable.sql
iriberri: Uops. I was not familiar at all with errors in postgres, but found: http://www.postgresql.org/docs/9.4/static/errcodes-appendix.html, so giving it a try. Thanks!

scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -466,19 +466,33 @@ END;
 $$ LANGUAGE 'plpgsql';
 
 
+
+DO $$
+BEGIN
+  IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = '_cdb_has_usable_primary_id_record') THEN
+    CREATE TYPE _cdb_has_usable_primary_id_record
```
rafatower: I would use types sparingly. Alternatively the function can return a `RECORD`

scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -499,40 +513,62 @@ BEGIN
...
-            RAISE DEBUG 'CDB(_CDB_Has_Usable_Primary_ID): %', Format('column %s contains nulls', const.pkey);
-            useable_key := false;
-          -- Other fatal error
-          WHEN others THEN
-            PERFORM _CDB_Error(sql, '_CDB_Has_Usable_Primary_ID');          
-        END;
-  
+
+        -- If the column is varchar, try to cast it to integer
+        IF rec.atttypid IN (1043) THEN
```
rafatower: the point here is not to run checks on the type of the column (and btw I really hate those magic numbers) but to assume the `cartodb_id` is valid just because you're feeding cartodbfy with it.

In other words: the contract of the `CDB_CartodbfyTable` should state that _if the input table contains a column `cartodb_id` it must be a valid one: numeric, unique and not null._ as a pre-condition.

scripts-available/CDB_CartodbfyTable.sql
iriberri: Agree :-) I had some issues late yesterday with the returns and one of my attempts was to force this, but it's not needed anymore.

scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -893,9 +932,16 @@ BEGIN
   -- We can only avoid a rewrite if both the key and 
   -- geometry are usable
 
+  -- If cartodb_id column is of type text, cast it
+  IF idc.text_key_column THEN
+    sql := Format('ALTER TABLE %s ALTER cartodb_id TYPE int USING %I::integer', reloid::text, 'cartodb_id');
```
rafatower: this is precisely what we're trying to avoid... I'll explain to you the whole thing (todo el chiringuito)

iriberri:
So:

* We're gonna asume that whatever `cartodb_id` we get is valid. If not, import failure.
* Due to this, we can just ignore the checks of type for the column cartodb_id and force in the Rewrite table function it's casting to big int. If cartodb_id doesn't follow the requirement this will just fail
* We can still maintain the int checks for columns not named `cartodb_id`. It might probably be worthy to ignore them and generate a new cartodb_id from scratch, but not the point right now

scripts-available/CDB_CartodbfyTable.sql
iriberri: Conclusions: https://github.com/CartoDB/cartodb-postgresql/pull/202#issuecomment-191180839

scripts-available/CDB_CartodbfyTable.sql
iriberri: After talking, this was totally me reading incorrectly the CDB Rewrite Table and getting confused by other alters.

iriberri:
Okay, if I didn't miss anything again, this is now more aligned.

I didn't edit at all the branch of "if there's another valid column not named cartodb_id..." as we said and as I wrote in my last comment.

test/CDB_CartodbfyTableTest_expect
```diff
@@ -28,6 +28,23 @@ trigger-protected the_geom cartodbfied fine
 extent|BOX(1 1,2 2)
 DROP TABLE
 SELECT 1
+text cartodb_id cartodbfied fine
+5
+DROP TABLE
+SELECT 1
+ERROR:  CDB(_CDB_Rewrite_Table:22P02:invalid input syntax for integer: "nan"): CREATE TABLE public.t_0 AS SELECT cartodb_id::bigint ,NULL::geometry(Geometry,4326) AS the_geom,NULL::geometry(Geometry,3857) AS the_geom_webmercator FROM t
```
iriberri: Wondering if I could make the CartoDBfy test function to catch these errors and output a less verbose error instead of performing tests against this specific context.

iriberri: Answering to myself: I set verbosity to tense to avoid contexts.

rafatower:
:+1: time to prepare a release

iriberri:
From @rafatower's feedback, add test:

`cartodbfy(t) = cartodbfy(cartodbfy(t))`

iriberri:
Some inconsistency here:

**Scenario A**:

````
CREATE TABLE table_50(
    cartodb_id    varchar,
    title         varchar(40)
);
INSERT INTO table_50 (cartodb_id) VALUES ('1');
INSERT INTO table_50 (cartodb_id) VALUES ('1');
select cartodb.cdb_cartodbfytable('table_50');
````
CartoDBfication works, result:

````
 cartodb_id | the_geom | the_geom_webmercator | cartodb_id_0 | title 
------------+----------+----------------------+--------------+-------
          1 |          |                      | 1            | 
          2 |          |                      | 1            | 
(2 rows)
````

**Scenario B**:


````
CREATE TABLE table_51(
    cartodb_id    varchar,
    title         varchar(40)
);
INSERT INTO table_51 (cartodb_id) VALUES ('a');
INSERT INTO table_51 (cartodb_id) VALUES ('b');
select cartodb.cdb_cartodbfytable('table_51');
````
CartoDBfication fails:

````
ERROR:  CDB(_CDB_Rewrite_Table:22P02:invalid input syntax for integer: "a"
````

Both cases are out of the requirements, but both are failing differently.

**Comment by iriberri** [comment]
The new commit makes failures consistent: everything will just crash now.

I've updated the tests too, so as soon as I fix the Rails part to give good feedback about the failure we're good.

**Comment by iriberri** [comment]
Tests passing :-) @rafatower could you give this a look again?

**Comment by rafatower** [comment]
:+1:

File: scripts-available/CDB_CartodbfyTable.sql [hunk 7]
```diff
@@ -479,6 +480,7 @@ DECLARE
   i INTEGER;
   sql TEXT;
   useable_key BOOLEAN = false;
+  has_usable_primary_key BOOLEAN = false;
```
iriberri: Unused variable.

iriberri:
My tests are going on here: https://gist.github.com/iriberri/d839f6e67e05c6454a1d

PR #204, opened 2016-03-02
Fully qualify invalidation calls (Fixes #198)

### LINKED ISSUE #198: Fully qualify table names on trigger invalidation caches
Right now, as we generate the invalidation tags based on CDB_QueryTablesText output, we tag tables with:
`schemaname.tablename`
even if the schema name could be omitted. This is because CDB_QueryTables formats it explicitly:
https://github.com/CartoDB/cartodb-postgresql/blob/master/scripts-available/CDB_QueryTables.sql#L48+

However, when we call cdb_invalidate_varnish, we pass it a casted to text `regclass`, which might only be `tablename` if `schemaname` is first on the search path.

This works _de chiripa_ as our regex is permissive enough on upper layers to invalidate X-Cache-Channel for `schemaname.tablename` when `tablename` is sent, but might break any invalidations which need strict matches such as Surrogate-Key's.

zenitraM:
We were relying on the explicit `regclass -> text` casting to send the message to invalidate a table. 
This means that a table on a schema on the search_path (i.e `public.foobar`) was being invalidated as `foobar`, but `CDB_QueryTables` tagged it as `public.foobar` (see desc in #198). 

This fixes this behavior by explicitly searching the table and schema name on the call to `cdb_invalidate_varnish` and properly formatting it.

cc @rafatower @rochoa

### DIFF
--- a/scripts-available/CDB_TableMetadata.sql
+++ b/scripts-available/CDB_TableMetadata.sql
@@ -66,9 +66,11 @@ CREATE OR REPLACE FUNCTION _CDB_TableMetadata_Updated()
 RETURNS trigger AS
 $$
 DECLARE
-  tabname TEXT;
+  tabname regclass;
   rec RECORD;
   found BOOL;
+  schema_name TEXT;
+  table_name TEXT;
 BEGIN
 
   IF TG_OP = 'UPDATE' or TG_OP = 'INSERT' THEN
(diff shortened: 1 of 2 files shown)


rafatower:
:+1:

rochoa:
LGTM.

Issue #209, opened 2016-03-15
CartoDB Rails tests expect cartodb_id not to be casted to bigint


iriberri:
but we casted it in: https://github.com/CartoDB/cartodb-postgresql/pull/202/files#diff-767fb5c7568f72bdf3da71fbe0d21ca4R922

This needs to be reverted.

PR #210, opened 2016-03-15
Remove casting to bigint in cartodb_id column

iriberri:
Fixes https://github.com/CartoDB/cartodb-postgresql/issues/209

We forced the casting to bigint thinking that every cartodb_id was created as a bigint, but it was not true and Rails tests made us notice.

For cartodbfycations of tables with an existent pk in cartodb_id, integer will be created, as everything is set up and manipulations on the column are not needed.

For cartodbfycations of tables without any cartodb_id at all, a bigint will be created due to the sequence that is assigned to it. `setval` and `nextval` return with bigints: http://www.postgresql.org/docs/9.4/static/functions-sequence.html

If we cast to `integer` instead, the support for text cartodb_id's will be still valid and it will not conflict with the legacy cartodb_id types.

### DIFF
--- a/scripts-available/CDB_CartodbfyTable.sql
+++ b/scripts-available/CDB_CartodbfyTable.sql
@@ -919,7 +919,7 @@ BEGIN
 
   -- Add cartodb ID!
   IF has_usable_primary_key THEN
-    sql := sql || const.pkey || '::bigint ';
+    sql := sql || const.pkey || '::integer ';
   ELSE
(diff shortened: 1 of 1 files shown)


iriberri:
@rafatower could you please take a look? :-)

rafatower:
:+1:

iriberri:
Manual tests (staging) and Rails tests (local) passed.

PR #215, opened 2016-03-21
Fix backslashes in generated files

### LINKED ISSUE #213: Weird Makefile problem
There's a problem with the doubled (escaped) backslahes here: https://github.com/CartoDB/cartodb-postgresql/blob/224b2ce39504f974cc47602fd49c3fecb60a5b27/Makefile#L126-L133

That was expected to generate single backslashes, but now in some systems it generates two backslahes. Actually the latter behaviour is most consistent, since it is the same result as executing the `echo` command directly in `bash`, not through `make`.

The only difference I've been able to find in the cases where the new behaviour (double slashes) occurs is that they're running bash 4.2.25 instead of 4.2.24. (Oddly enough, OSX with bash 3.2.25 behaves as 4.2.24 even though it includes the same patch as 4.2.25)

jgoizueta:
Fixes #213
Some systems treat escaping of text passed to the echo command
from a Makefile differently.
This seems to work for our needs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,6 +116,9 @@ $(EXTENSION).control: $(EXTENSION).control.in Makefile
 cartodb_version.sql: cartodb_version.sql.in Makefile $(GITDIR)/index
 	$(SED) -e 's/@@VERSION@@/$(EXTVERSION)/' $< > $@
 
+# Needed for consistent `echo` results with backslashes
+SHELL = bash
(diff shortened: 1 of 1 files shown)


jgoizueta:
Please @rafatower @ethervoid  CR this

rafatower:
:+1

Issue #218, opened 2016-03-30
CDB_Overviews(reloid REGCLASS) fails for org users


rafatower:
```
=> SELECT cartodb.CDB_DropOverviews('rtorre.yellow_tripdata_2015_pickups'::regclass);
ERROR:  relation "_vovw_19_yellow_tripdata_2015_pickups" does not exist
CONTEXT:  SQL function "cdb_overviews" statement 1
PL/pgSQL function cdb_dropoverviews(regclass) line 5 at FOR over SELECT rows
```

```
=> \dt rtorre.*yellow*                                                                                                                                                                                                                  List of relations
 Schema |                 Name                  | Type  |                       Owner                       
--------+---------------------------------------+-------+---------------------------------------------------
 rtorre | _vovw_0_yellow_tripdata_2015_pickups  | table | cartodb_user_uuid-lalala
 rtorre | _vovw_10_yellow_tripdata_2015_pickups | table | cartodb_user_uuid-lalala
 rtorre | _vovw_11_yellow_tripdata_2015_pickups | table | cartodb_user_uuid-lalala
 rtorre | _vovw_12_yellow_tripdata_2015_pickups | table | cartodb_user_uuid-lalala
 rtorre | _vovw_13_yellow_tripdata_2015_pickups | table | cartodb_user_uuid-lalala
 rtorre | _vovw_14_yellow_tripdata_2015_pickups | table | cartodb_user_uuid-lalala
 rtorre | _vovw_15_yellow_tripdata_2015_pickups | table | cartodb_user_uuid-lalala
 rtorre | _vovw_16_yellow_tripdata_2015_pickups | table | cartodb_user_uuid-lalala
 rtorre | _vovw_17_yellow_tripdata_2015_pickups | table | cartodb_user_uuid-lalala
 rtorre | _vovw_18_yellow_tripdata_2015_pickups | table | cartodb_user_uuid-lalala
 rtorre | _vovw_19_yellow_tripdata_2015_pickups | table | cartodb_user_uuid-lalala
 rtorre | _vovw_1_yellow_tripdata_2015_pickups  | table | cartodb_user_uuid-lalala
 rtorre | _vovw_2_yellow_tripdata_2015_pickups  | table | cartodb_user_uuid-lalala
 rtorre | _vovw_3_yellow_tripdata_2015_pickups  | table | cartodb_user_uuid-lalala
[...]

PR #224, opened 2016-04-06
Fix CDB_DropOverviews and CDB_Overviews problems

### LINKED ISSUE #223: CDB_DropOverviews fails for org users
This bug has been introduced in 0.15.0: https://github.com/CartoDB/cartodb-postgresql/blob/e68d5eca4587d17da4eee9ade260fb4c7eed32c4/scripts-available/CDB_Overviews.sql#L123

Since `row.overview_table` is a regclass it should not be formatted with '%I' and it should not be prependend a schema. The problem will manifest when the overview_table requires the schema 
name to be accessed (i.e. when casting the regclass to text contains the schema part).

The bug is not detected by tests because no tables in schemas that are not in the `search_path` are used in the tests.

jgoizueta:
Fixes #223 and #218

### DIFF
--- a/scripts-available/CDB_Overviews.sql
+++ b/scripts-available/CDB_Overviews.sql
@@ -370,7 +370,7 @@ AS $$
 
     SELECT * FROM _cdb_split_table_name(reloid) INTO schema_name, table_name;
 
-    EXECUTE Format('DROP TABLE IF EXISTS %I.%I CASCADE;', schema_name.overview_rel);
+    EXECUTE Format('DROP TABLE IF EXISTS %I.%I CASCADE;', schema_name, overview_rel);
 
     -- Estimate number of rows
     SELECT reltuples, relpages FROM pg_class INTO STRICT class_info
--- a/scripts-available/CDB_Overviews.sql
+++ b/scripts-available/CDB_Overviews.sql
@@ -120,7 +120,7 @@ BEGIN
     FOR row IN
         SELECT * FROM CDB_Overviews(reloid)
     LOOP
-        EXECUTE Format('DROP TABLE %I.%I;', schema_name, row.overview_table);
+        EXECUTE Format('DROP TABLE %s;', row.overview_table);
         RAISE NOTICE 'Dropped overview for level %: %', row.z, row.overview_table;
     END LOOP;
 END;
--- a/scripts-available/CDB_Overviews.sql
+++ b/scripts-available/CDB_Overviews.sql
@@ -221,7 +221,7 @@ AS $$
 
     BEGIN
       EXECUTE ext_query INTO ext;
-      EXCEPTION
+    EXCEPTION
         -- This is the typical ERROR: stats for "mytable" do not exist
         WHEN internal_error THEN
           -- Get stats and execute again
--- a/scripts-available/CDB_Overviews.sql
+++ b/scripts-available/CDB_Overviews.sql
@@ -142,9 +142,12 @@ AS $$
     schema_name TEXT;
     table_name TEXT;
   BEGIN
-    -- TODO: review implementation of CDB_UserTables an suitability for this
     SELECT * FROM _cdb_split_table_name(reloid) INTO schema_name, table_name;
-    RETURN QUERY SELECT
+    -- TODO: replace use of CDB_UserTables by obtaining the user tables
+    --       in a specific schema
+    -- Meanwhile we'll use DISTINCT here to avoid picking multiple tables
+    -- from different schemas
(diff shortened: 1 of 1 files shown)


jgoizueta:
Please @rafatower CR this. We can release this as 0.15.1 to solve all the organization-user-related problems I've detected.

jgoizueta:
cc @ethervoid

scripts-available/CDB_Overviews.sql
```diff
@@ -1,4 +1,30 @@
...
+  FROM pg_class c
+  JOIN pg_namespace n ON n.oid = c.relnamespace
+  WHERE c.relkind = 'r'
+  AND c.relname NOT IN ('cdb_tablemetadata', 'spatial_ref_sys')
+  AND CASE WHEN schema_name IS NULL
+             THEN n.nspname NOT IN ('pg_catalog', 'information_schema', 'topology', 'cartodb')
+           ELSE n.nspname = schema_name END
+  AND CASE WHEN perm = 'public' THEN has_table_privilege('publicuser', c.oid, 'SELECT')
+           WHEN perm = 'private' THEN has_table_privilege(current_user, c.oid, 'SELECT') AND NOT has_table_privilege('publicuser', c.oid, 'SELECT')
+           WHEN perm = 'all' THEN has_table_privilege(current_user, c.oid, 'SELECT') OR has_table_privilege('publicuser', c.oid, 'SELECT')
```
rafatower: Minor comment: 

I understand that this is basiclally copied from CDB_UserTables with some tweaks. But I also understand that if the user has perms on the base table then they should be able to perform any operation on them regardless of the perms in overviews tables.

Therefore the code to check privileges could be removed? the less code the less maintenance.

But as it is tested, ok, go ahead with it :)

scripts-available/CDB_Overviews.sql
```diff
@@ -140,16 +166,15 @@ RETURNS TABLE(base_table REGCLASS, z integer, overview_table REGCLASS)
...
-    -- TODO: review implementation of CDB_UserTables an suitability for this
-    SELECT * FROM _cdb_split_table_name(reloid) INTO schema_name, table_name;
+    SELECT * FROM _cdb_split_table_name(reloid) INTO schema_name, _table_name;
     RETURN QUERY SELECT
       reloid AS base_table,
-      _CDB_OverviewTableZ(cdb_usertables) AS z,
-      ('"' || schema_name|| '"."' ||cdb_usertables || '"')::regclass AS overview_table
-      FROM CDB_UserTables()
-      WHERE _CDB_IsOverviewTableOf((SELECT relname FROM pg_class WHERE oid=reloid), cdb_usertables)
+      _CDB_OverviewTableZ(table_name) AS z,
```
rafatower: there's a `table_name` and a `_table_name` (prefixed with underscore). Don't know when the one w/o underscore is declared.

if it is declared, they have bad naming making the code harder to understand

scripts-available/CDB_Overviews.sql
```diff
@@ -204,7 +231,7 @@ AS $$
         -- This is the typical ERROR: stats for "mytable" do not exist
         WHEN internal_error THEN
           -- Get stats and execute again
-          EXECUTE format('ANALYZE %1$I', reloid);
+          EXECUTE format('ANALYZE %1$s', reloid);
```
rafatower: what's this ANALYZE for?

scripts-available/CDB_Overviews.sql
jgoizueta: Not a bug but confusing naming: the one with underscore is variable used to get the second result of `_cdb_split_table_name` which isn't actually used; the other identifer is a column of `_CDB_UserTablesInSchema`. Should change the name of the first to make more evident it not being used.

scripts-available/CDB_Overviews.sql
jgoizueta: I agree to remove it

scripts-available/CDB_Overviews.sql
jgoizueta: I hoped you knew! (you placed that code there). Is it supposed to generate the stats, the lack of which could have provoked an error?

rafatower: :dancer: now I remember: when using the quick extend function, there's little chance of error if it doesn't have the stats. unlikely to happen.

iriberri:
@jgoizueta I just pushed the changes for the bugs we found while testing. I a brief description of them in the commit description.

jgoizueta:
:+1:

Issue #226, opened 2016-04-12
Update Overviews queries to get the count of the aggregated features


iriberri:
cc @jgoizueta

iriberri:
This change will break the tiler as it's including a new column. The changes in the tiler will be performed from this issue: https://github.com/CartoDB/Windshaft-cartodb/issues/413

iriberri:
We have changed the strategy and we no longer need the tiler to add the extra column because we will actually add it to the base table.

PR #227, opened 2016-04-12
Include and aggregate _vovw_count column to count aggregated features

iriberri:
Closes https://github.com/CartoDB/cartodb-postgresql/issues/226

cc @jgoizueta

### DIFF
--- a/scripts-available/CDB_Overviews.sql
+++ b/scripts-available/CDB_Overviews.sql
@@ -718,6 +718,7 @@ DECLARE
   overview_z integer;
   overview_tables REGCLASS[];
   overviews_step integer := 1;
+  has_counter_column boolean;
 BEGIN
   -- Determine the referece zoom level
   EXECUTE 'SELECT ' || quote_ident(refscale_strategy::text) || Format('(''%s'', %s);', reloid, tolerance_px) INTO ref_z;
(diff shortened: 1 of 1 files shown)


jgoizueta:
👍

iriberri:
Hey @ethervoid could you take a look? :-)

ethervoid:
Despite I don't know much about how it works, it looks good to me

jgoizueta:
The *ugly* `_vovw_count` name has been chosen because all internal details related to overviews are named with a `_vovw_` prefix (the underscore to avoid collision with user identifiers, the vovw meaning "vector overviews").

The columns counts the number of features aggregated (clustered?)  into a single overview feature.

Any suggestion to change this name to something nicer?

cc @javisantana

javisantana:
`point_count` would be better but please, don't use obfuscated names

javisantana:
there is one thing, what happen is there is already a column `_vovw_count` ? (or whatever is the name)

jgoizueta:
Currently, if there's an existing `_vovw_count` we expect it to comply with our requirements (be a numeric  column with constant value 1), similarly to what we do with `cartodb`, `the_geom`.

For tables the `_vovw` prefix prevents collision with user tables because we don't allow user tables' names to start with underscore, but that's possible for user columns, sadly.

If we use `point_count` likehood of collision would be much higher. Also we may eventually have overviews por polygons...

What about  `_feature_count`?

javisantana:
sounds good for me

iriberri:
Sounds good to me too. Renamed :+1:.

jgoizueta:
:+1:

PR #228, opened 2016-04-13
Use cell centers, not cluster centroids when grouping points

### LINKED ISSUE #225: Make the overviews point clustering more like torque
Currently points are clustered using a grid (like torque does), but the aggregated point is placed at the centroid of each cluster rather than the cell's center like torque does. This seems to create some artifacts. This is an overviews layer:

ovw1

And this is a torque layer:

ovw2

jgoizueta:
Fixes #225

### DIFF
--- a/scripts-available/CDB_Overviews.sql
+++ b/scripts-available/CDB_Overviews.sql
@@ -596,6 +596,7 @@ AS $$
     gtypes TEXT[];
     schema_name TEXT;
     table_name TEXT;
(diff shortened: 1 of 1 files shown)


jgoizueta:
cc @iriberri @rafatower please review

iriberri:
LGTM. In the other PR I added a couple of small fixes to that function: setting the grid to 1.0 by default and deleting a duplicated RETURN, :+1:

Issue #229, opened 2016-04-14
Set the default overviews tolerance in pixels to 1


jgoizueta:
If we use an integral tolerance value, the points in overviews are gridded regularly on pixels, but if 
TWKB is used by the tiler, the rounding of coordinates causes irregularities in the grid at the pixel level. (the `twkb_rounding_adjustment` parameter of the tiler can be used with a negative value for finer rounding making the effect less frequent, but the effect cannot be avoided completely).

The effect is most noticeable with the current default value of 2 if 1-pixel-wide markers are used. Making the default 1 pixel alleviates this problem and also enhances the appearance of overviews, areas densely packed with features appear to fill the space even with 1-pixel markers.

The 1-pixel grid also makes the maximum number of features per tile to be 65535 which is the value we have been using in some configurations.

PR #230, opened 2016-04-14
Set tolerance to 1 pixel in overviews by default

iriberri:
Closes https://github.com/CartoDB/cartodb-postgresql/issues/229


jgoizueta:
👍

PR #233, opened 2016-04-15
Fix AVG computation in overview tables

### LINKED ISSUE #232: Overviews AVG aggregation is not correct
Since overviews are generated in cascade, averages of lower Z levels are averages of averages (which is incorrect)

jgoizueta:
Fixes #232
Averages of averages are not equal to overall averages.


jgoizueta:
please @iriberri CR this

iriberri:
:+1:

PR #234, opened 2016-04-15
Aggregate small number of text items in overviews

### LINKED ISSUE #231: Better text aggregation for overviews
Overview features representing more than one base feature have their text columns nulled.

It would be useful (e.g. for hover infowindows) to show the multiple text values aggregated when they are few and short.

jgoizueta:
Instead of nulling text fields for non-singleton aggregated records
concatenate distinct text values when they're few (5 or less).
Fixes #231

### DIFF
--- a/scripts-available/CDB_Overviews.sql
+++ b/scripts-available/CDB_Overviews.sql
@@ -549,7 +549,6 @@ DECLARE
   has_counter_column BOOLEAN;
   feature_count TEXT;
   total_feature_count TEXT;
-  unlimited_text BOOLEAN;
 BEGIN
   IF table_alias <> '' THEN
     qualified_column := Format('%I.%I', table_alias, column_name);
@@ -578,8 +577,7 @@ BEGIN
       RETURN Format('SUM(%s*%s)/%s::' || column_type, qualified_column, feature_count, total_feature_count);
     END IF;
   WHEN 'text', 'character varying', 'character' THEN
-    SELECT _cdb_unlimited_text_column(reloid, column_name) INTO unlimited_text;
-    IF unlimited_text THEN
+    IF _cdb_unlimited_text_column(reloid, column_name) THEN
       -- TODO: this should not be applied to columns containing largish text;
       -- it is intended only to short names/identifiers
       RETURN  'CASE WHEN count(distinct ' || qualified_column || ') = 1 THEN MIN(' || qualified_column || ') WHEN ' || total_feature_count || ' < 5 THEN string_agg(distinct ' || qualified_column || ','' / '') ELSE ''*'' END::' || column_type;
--- a/scripts-available/CDB_Overviews.sql
+++ b/scripts-available/CDB_Overviews.sql
@@ -515,6 +515,23 @@ BEGIN
 END
 $$ LANGUAGE PLPGSQL STABLE;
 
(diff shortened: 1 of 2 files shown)


jgoizueta:
cc @rafatower we're think we can merge this and release 0.15.1 (all other changes are already merged).

This last change has been succesfully tested with the "large" tables (including the 13.8M). It provides better aggregated results for the text fields, which is specially nice for 'hovering' use: when a point aggregate few (5 or less) items, text fields are concatenated; otherwise a '*' is shown (we've checked having the asterisk instead of NULL doesn't significantly increase table size).

scripts-available/CDB_Overviews.sql
```diff
@@ -559,16 +577,15 @@ BEGIN
...
-    -- e.g. with '/' as separator and a limit of three:
-    --     'A', 'B', 'A', 'C', 'D' => 'A/B/C/...'
-    -- Other ideas: if value is unique then use it, otherwise use something
-    -- like '*' or '(varies)' or '(multiple values)', or NULL
-    -- Using 'string_agg(' || qualified_column || ',''/'')'
-    -- here causes
-    RETURN 'CASE count(*) WHEN 1 THEN MIN(' || qualified_column || ') ELSE NULL END::' || column_type;
+  WHEN 'text', 'character varying', 'character' THEN
+    SELECT _cdb_unlimited_text_column(reloid, column_name) INTO unlimited_text;
+    IF unlimited_text THEN
```
rafatower: I think in this kind of expression you can simply do:
```
IF _cdb_unlimited_text_column(reloid, column_name) THEN
-- [...]
END IF;
```

scripts-available/CDB_Overviews.sql
```diff
@@ -559,16 +577,15 @@ BEGIN
...
+      -- TODO: this should not be applied to columns containing largish text;
+      -- it is intended only to short names/identifiers
+      RETURN  'CASE WHEN count(*) = 1 THEN MIN(' || qualified_column || ') WHEN ' || total_feature_count || ' < 5 THEN string_agg(distinct ' || qualified_column || ','' / '') ELSE ''*'' END::' || column_type;
```
rafatower: shouldn't it be `count(distinct(qualified_column))` instead of `count(*)`?

jgoizueta: I have to check if that works and is efficient, because it would be a great improvement.
(I fear it may affect performance, specially for multiple text columns, because it has to count distinct values of each text column per group, instead of just using the group size).

rafatower:
okay, I refrained myself from commenting on performance but now that you mention it... think how important this is to have for this milestone and how risky is adding it.

I mean (questions that I'd pose to myself if I were you):
- is this a nice-to-have or something we absolutely need in order to release?
- if I'm afraid of performance issues: how much time would I need to spend testing to be confident of deploying it?
- if I found a performance issue: how much time would I have to spend on optimizing it, if possible?

In the light of those questions, my answer would probably be: if it is not a must-have for closing with counters, leave it for a future iteration.

But I'll let you decide :)

scripts-available/CDB_Overviews.sql
jgoizueta: Right, I got confused with some similar case I handled before

PR #236, opened 2016-04-18
Log Bins & StdDev Bins

AbelVM:
## Log bins

Simple function to calculate logarithmic (base 10) bucketing for a data field.

The third parameter defines the point of interest where the bucketing is going to focus on:
- **start:** the lowest value of the input
- **end:** the highest value of the input
- **center:** the center of the distribution
- **''** (empty string): the poi will be guessed, using the bucketed mode
- **'N'** (numerical value): an arbitrary numeric value.
## StdDev bins

Standard deviation bucketing, the number of bins is calculated, so no need to set it


javisantana:
Hey, 

we need to work on clustering methods but doing them in the right way (the postgres way). I know it's harder but in the way we ensure it's maintained properly and has the required level of quality.

Algunenano:
Stale

PR #237, opened 2016-04-18
Adds table for storing camshaft analysis nodes

rochoa:


rochoa:
cc @javisantana @dgaubert Do you mind having a look? 🙏

dgaubert:
👍

rochoa:
cc @zenitraM

scripts-available/CDB_AnalysisCatalog.sql
```diff
@@ -0,0 +1,24 @@
+-- Table to register analysis nodes from https://github.com/cartodb/camshaft
+CREATE TABLE IF NOT EXISTS
+cartodb.cdb_analysis_catalog (
+    -- md5 hex hash
+    node_id char(40) CONSTRAINT cdb_analysis_catalog_pkey PRIMARY KEY,
+    -- being json allows to do queries like analysis_def->>'type' = 'buffer'
+    analysis_def json NOT NULL,
+    -- can reference other nodes in this very same table, allowing recursive queries
+    input_nodes char(40) ARRAY NOT NULL DEFAULT '{}',
```
zenitraM: would it make sense (or be possible) to add a foreign key here? so you cannot drop parent input_nodes without dropping their outputs

zenitraM:
:+1:

scripts-available/CDB_AnalysisCatalog.sql
rochoa: Oh, nice suggestion! I just looked if that was possible but unfortunately postgresql does not support that yet. And I wouldn't want to create more tables for storing the relations.

PR #238, opened 2016-04-18
Optimize column information functions

### LINKED ISSUE #235: Overview creation is too slow for tables with many columns
Generating the aggregation expression for the columns of a table is too slow; e.g. it may take about 50sec for 36 columns. The calls to `CDB_ColumnNames` and `_cdb_unlimited_text_column` from `__CDB_Attribute_Aggregation_Expression` seem to be the main culprit (removing them reduces the execution time of this function to less than a fifth).

All the column information needed could be obtained in a single query out of the columns iteration, then provided to the `_CDB_Attribute_Aggregation_Expression` function.

jgoizueta:
Fixes #235

### DIFF
--- a/scripts-available/CDB_ColumnNames.sql
+++ b/scripts-available/CDB_ColumnNames.sql
@@ -5,9 +5,9 @@ AS $$
   SELECT
     a.attname::information_schema.sql_identifier column_name
     FROM pg_class c
-         LEFT JOIN pg_catalog.pg_attribute a ON a.attrelid = c.oid
+         LEFT JOIN pg_attribute a ON a.attrelid = c.oid
     WHERE c.oid = $1::oid
-    AND a.attstattarget = -1
+    AND a.attstattarget < 0 -- exclude system columns
    ORDER BY a.attnum;
 $$ LANGUAGE SQL;
 
--- a/scripts-available/CDB_ColumnNames.sql
+++ b/scripts-available/CDB_ColumnNames.sql
@@ -2,15 +2,13 @@
 CREATE OR REPLACE FUNCTION CDB_ColumnNames(REGCLASS)
 RETURNS SETOF information_schema.sql_identifier
 AS $$
-
-    SELECT c.column_name
-      FROM information_schema.columns c, pg_class _tn, pg_namespace _sn
-      WHERE table_name = _tn.relname
-        AND table_schema = _sn.nspname
-        AND _tn.oid = $1::oid
-        AND _sn.oid = _tn.relnamespace
-      ORDER BY ordinal_position;
-
+  SELECT
+    a.attname::information_schema.sql_identifier column_name
+    FROM pg_class c
+         LEFT JOIN pg_catalog.pg_attribute a ON a.attrelid = c.oid
+    WHERE c.oid = $1::oid
+    AND a.attstattarget = -1
+   ORDER BY a.attnum;
 $$ LANGUAGE SQL;
 
 -- This is to migrate from pre-0.2.0 version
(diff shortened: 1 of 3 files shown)


jgoizueta:
Please @rafatower review this

scripts-available/CDB_ColumnNames.sql
```diff
@@ -2,15 +2,13 @@
...
-      WHERE table_name = _tn.relname
-        AND table_schema = _sn.nspname
-        AND _tn.oid = $1::oid
-        AND _sn.oid = _tn.relnamespace
-      ORDER BY ordinal_position;
-
+  SELECT
+    a.attname::information_schema.sql_identifier column_name
+    FROM pg_class c
+         LEFT JOIN pg_catalog.pg_attribute a ON a.attrelid = c.oid
```
rafatower: I'd take advantage of the change to use better names, such as `attribute` instead of `a`

scripts-available/CDB_ColumnNames.sql
```diff
@@ -2,15 +2,13 @@
...
+    WHERE c.oid = $1::oid
+    AND a.attstattarget = -1
```
rafatower: I don't understand the usage of `attstattarget` and the magic number `-1`

http://www.postgresql.org/docs/9.4/static/catalog-pg-attribute.html

what am I missing?

scripts-available/CDB_Overviews.sql
```diff
@@ -520,15 +520,14 @@ CREATE OR REPLACE FUNCTION _cdb_unlimited_text_column(reloid REGCLASS, col_name
...
-        AND character_maximum_length IS NULL
-        AND c.data_type IN ('text', 'character varying', 'character')
+    SELECT a.attname
+    FROM pg_class c
+         LEFT JOIN pg_catalog.pg_attribute a ON a.attrelid = c.oid
+         LEFT JOIN pg_catalog.pg_type t ON t.oid = a.atttypid
+    WHERE c.oid = reloid
+      AND a.attname = col_name
+      AND pg_catalog.format_type(a.atttypid, NULL) IN ('text', 'character varying', 'character')
+      AND pg_catalog.format_type(a.atttypid, NULL) = pg_catalog.format_type(a.atttypid, a.atttypmod)
```
rafatower: AFAIK the `pg_catalog` is always implicitly in the search_path. Did you have trouble with it?

rafatower: I don't get this line or the why of the textual comparison of types.

scripts-available/CDB_ColumnNames.sql
jgoizueta: Regular attributes (user columns) have -1 here (which indicates stats should be collected). This is to ignore internal attributes such as oid

scripts-available/CDB_Overviews.sql
jgoizueta: Looking at it I actually don't remember how I came to that... I guess the literal type names use is just due to laziness 😓 . I'll rewrite that without the explicit pg_catalog and will se type ids...

rafatower: okay, so I see my question made some little sense :smile: I didn't mean you to change it, just understand the code.

If it works as intended you can just create a ticket to remember and go ahead. you know, something that works today is better than something "perfect" tomorrow. :wink:

scripts-available/CDB_ColumnNames.sql
rafatower: okay, as long as you are 100% confident of what you are doing (as I cannot find such information and it seems to be an internal implementation detail that may change in future versions)

PR #242, opened 2016-04-25
Adjust overview points to pixel centers

### LINKED ISSUE #240: Overview clusters should be adusted to pixel centers
Overviews can be generated with a `tolerance_px` parameter that clusters points to a pixel-based grid based on the webmercator tiles that the tiler (Windshaft) renders.

When TWKB is used by the tiler (to minimize the size of the data transfered from the database)
the coordinates are rounded to some number of decimals that by default are enough to address
each pixels uniquely (the least-significant-digit of the rounded coordinates are smaller than a pixel;
but this can be controlled by the [`twkb_rounding_adjustment`](https://github.com/CartoDB/mapnik/tree/2.3.x.cartodb/plugins/input/postgis#23xcartodb) parameter).

If the pixel-adjusted coordinates are not computed at the center of pixels, artifacts as in the image (specially noticeable when the grid's size is 2 pixels) can occur even when the TWKB coordinates have subpixel precision:

tile_14_4824_6157_bad

These artefacts can be eliminated by forcing the point coordinates (before TWKB’s rounding) to lie at pixel’s centres. This would work with TWKB for any `twkb_rounding_adjustment` less than 0.5.

Currently coordinates were computed at the cell’s centres; for for cells made up of an exact, odd number of pixels (`tolerance_px`) the point was already at a pixel’s centre and no artefacts were produced; for cells made up of an exact, even number of pixels the points lied in between pixels and the effect was most noticeable.

jgoizueta:
This fixes #240 

With these changes we eliminate pixel level artifacts of overviews in the presence of TWKB for any value of `twkb_rounding_adjustment` less than 0.5, including the default value of 0.

### DIFF
--- a/scripts-available/CDB_Overviews.sql
+++ b/scripts-available/CDB_Overviews.sql
@@ -655,6 +655,10 @@ AS $$
     pixel_m FLOAT8;
     grid_m FLOAT8;
     offset_m FLOAT8;
+    offset_x TEXT;
+    offset_y TEXT;
+    cell_x TEXT;
+    cell_y TEXT;
     aggr_attributes TEXT;
     attributes TEXT;
     columns TEXT;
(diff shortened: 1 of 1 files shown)


jgoizueta:
Please review, @rafatower 

cc @javisantana

rafatower:
nice! :+1: :_-)

PR #243, opened 2016-04-25
Compute webmercator resolution with full accuracy

### LINKED ISSUE #241: Full accuracy webmercator extents
The actual value of the webmercator equatorial circumference (used for tiling) is based on the value 6378137 for the Earth radiut (taken from the semimajor axis of WGS84’s ellipsoid). So, the size of the level 0 tile (2·R·PI) should be more accurately computed as 40075016.68557849 (current value is 40075017.0).

We should replace the definition of [CDB_XYZ_Resolution](https://github.com/CartoDB/cartodb-postgresql/blob/897689dd439e40256c4a66d02d78a7f31eaaf012/scripts-available/CDB_XYZ.sql#L9) with `SELECT 6378137.0*2.0*pi() / 256.0 / power(2.0, z);`  to get maximum accuracy of tile/pixel coordinates.

This will make the resolution values consistent with those used in the tiler:
https://github.com/CartoDB/tilelive-mapnik/blob/d04792444eb356d1e240ecd792f5d0ab933b213e/lib/render.js#L7

jgoizueta:
Fixes #241

### DIFF
--- a/scripts-available/CDB_XYZ.sql
+++ b/scripts-available/CDB_XYZ.sql
@@ -6,7 +6,7 @@ CREATE OR REPLACE FUNCTION CDB_XYZ_Resolution(z INTEGER)
 RETURNS FLOAT8
 AS $$
   -- circumference divided by 256 is z0 resolution, then divide by 2^z
-  SELECT 40075017.0 / 256 / power(2, z);
+  SELECT 6378137.0*2.0*pi() / 256.0 / power(2.0, z);
 $$ LANGUAGE SQL IMMUTABLE STRICT;
 -- }
 


jgoizueta:
Please review @rafatower

rafatower:
okay, but two questions:
- how this affects actual calculations? I mean, the pixel resolution is not the same in both axis, so an error is granted depending on the zoom level, right? (from my ignorance).
- have you tested what happens with the IMMUTABLE modifier when you update the extension? Will it actually take the newest, more accurate version of the function's results?

jgoizueta:
Basically we're adding a small (irrelevant for current applications) bun unnecessary error to the inevitable error due to floating point's limited precision. Not a big deal, but this makes the exact values computed inconsistent with those used in the tiler: https://github.com/CartoDB/tilelive-mapnik/blob/d04792444eb356d1e240ecd792f5d0ab933b213e/lib/render.js#L7

Regarding the immutability, I've checked that when the function is replaced the new values are used (even for z values previously computed)

rafatower:
:+1:

Issue #247, opened 2016-04-27
Nulls in binning functions


pnorman:
Right now our treatment of NULLs in binning functions doesn't make sense and is undocumented. As an example,

``` sql
SELECT cdb_quantilebins('{1,2,3,NULL,6,NULL,9}'::numeric[],4); -- returns {2,6,NULL,9}
SELECT cdb_quantilebins('{1,1,2,2,3,3,NULL,NULL,6,6,NULL,NULL,9,9}'::numeric[],4); -- returns {2,6.0000000000000000,NULL,9}
SELECT cdb_quantilebins('{1,2,3,6,9}'::numeric[],4); -- returns {2,3,6,9}
SELECT cdb_quantilebins('{1,2,3,6,9}'::numeric[],3); -- returns {2,6,9}
```

It is neither nulls first or nulls last, but nulls somewhere in the middle.

pnorman:
The [PostgreSQL docs](http://www.postgresql.org/docs/current/static/functions-aggregate.html#FUNCTIONS-ORDEREDSET-TABLE) state that "All the aggregates listed in Table 9-51 ignore null values in their sorted input." so I suggest we do the same.

Looking at the internals of CDB_QuantileBins the reason the NULL gets in the middle is it pulls from the sorted array for the 2nd last output element, which is a null, but the very last element is `max(in_array)` which ignores nulls

```
{1,2,3,6,9,NULL,NULL}
   ^   ^   ^
  #1  #2  #3
   
#4 = max = 9
```

Algunenano:
Closing this as it's no longer the case.

PR #249, opened 2016-04-28
Test behavior of binning fuctions with nulls

pnorman:
All test results are based off of existing behavior, which doesn't always make sense (ref #247), but I wanted to get tests in **before** considering anything that would change expected output.


pnorman:
cc @zenitraM

PR #250, opened 2016-04-29
Stats functions rewrite

pnorman:
- Switches to much shorter SQL implementations
- Fixes a bug with the wrong standard deviation being used
- Decreases tolerance for error on the tests
- Adds versions for all number types, not just numeric.

There is a minor (~10%) performance gain from the SQL version, but on a table with reals the new versions are over 4x faster.


Algunenano:
Stale

PR #253, opened 2016-05-09
Do not quote arguments to ST_EstimatedExtent

### LINKED ISSUE #252: Arguments to ST_EstimateExtent must not be quoted
`_cdb_estimated_extent` generates a call to `ST_EstimateExtent` quoting (with double quotes) its arguments when they need it as identifiers, but that causes a syntax error when `ST_EstimateExtent` evaluates the arguments

jgoizueta:
Fixes #252

### DIFF
--- a/scripts-available/CDB_Overviews.sql
+++ b/scripts-available/CDB_Overviews.sql
@@ -235,7 +235,7 @@ AS $$
       FROM pg_class c JOIN pg_namespace n on n.oid = c.relnamespace WHERE c.oid = reloid::oid;
 
     ext_query = format(
-      'SELECT ST_EstimatedExtent(''%1$I'', ''%2$I'', ''%3$I'');',
+      'SELECT ST_EstimatedExtent(''%1$s'', ''%2$s'', ''%3$s'');',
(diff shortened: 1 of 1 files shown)


jgoizueta:
@rafatower  @ethervoid  please take a look

scripts-available/CDB_Overviews.sql
```diff
@@ -235,7 +235,7 @@ AS $$
       FROM pg_class c JOIN pg_namespace n on n.oid = c.relnamespace WHERE c.oid = reloid::oid;
 
     ext_query = format(
-      'SELECT ST_EstimatedExtent(''%1$I'', ''%2$I'', ''%3$I'');',
+      'SELECT ST_EstimatedExtent(''%1$s'', ''%2$s'', ''%3$s'');',
```
rafatower: Wondering why a dynamic query is needed here. Can't you just use `RETURN QUERY`?

rafatower: never mind, if it works go ahead

rafatower:
:+1:

ethervoid:
Seems good to me

Issue #257, opened 2016-05-24
Should limit the maximum level for which to create overviews


jgoizueta:
If a table's points are all within a very small area, overviews will be created for high Z values that can cause overflow errors:

```
ERROR:  integer out of range
CONTEXT:  SQL statement "
      CREATE TABLE _vovw_31_chk AS
         WITH clusters AS (
           SELECT
             SUM(longitude*1)/count(*)::double precision AS longitude,SUM(latitude*1)/count(*)::double precision AS latitude,
             count(*) AS n,
             Floor(ST_X(f.the_geom_webmercator)/7.28960306979907e-05)::int AS gx,
             Floor(ST_Y(f.the_geom_webmercator)/7.28960306979907e-05)::int AS gy,
             MIN(cartodb_id) AS cartodb_id
          FROM jgoizueta.chk f
          GROUP BY gx, gy
         )
...
```

Here, trying to create overviews for level 31 will cause integer overflow because the clustering integer coordinates (`gx`, `gy`) would require more than 32 bits.

Since `gx` and `gy` are usually pixel clusters, they need 8 bits (for the 256 pixels per tile) more than what is needed for tile coordinate values (z+1 bits for zoom level z; the "+1" if to allow using signed integers)

This means if we keep on using integers for gx, gy we shouldn't create overviews for levels over 23.
If we upgrade gx, gy (and possible other expressions) to be `bigint's, we could theoretically reach level 55, but shouldn't go over 31 to avoid requiring more than 32 bits for the tile coordinates.

So we should either limit the zoom level to a maximum of **23** or use big integers in someplaces to allow overviews up to level **30**.

jgoizueta:
Note: the table that causes this problem was created with this Ruby snippet:

```ruby
puts "cartodb_id,longitude,latitude"
x0 = -3.0
y0 = 42.0
d = 1e-6
id = 0
10000.times do
  id += 1
  x = x0 + d*rand - d/2
  y = y0 + d*rand - d/2
  puts "#{id},#{x},#{y}"
end
```

PR #258, opened 2016-05-24
Fix problems computing feature density for overviews creation

### LINKED ISSUE #256: Error during feature density evaluation for overviews creation
In some cases, during overview creation an error occurs in the `_CDB_Feature_Density` function due to some expressions which should be integers being taken as floating point:

```
Exception: PG::Error: ERROR:  function cdb_xyz_extent(double precision, double precision, integer) does not exist
LINE 27:           WHERE the_geom_webmercator && CDB_XYZ_Extent(x*2 +...
                                                 ^
```

This has happened three times in production, but I haven't be able to reproduce so far.

jgoizueta:
We still have to determine if the fix fo #257 fixes #256 too (in all the cases I've tried the error is like in the first problem: integer too large, and not like the second: casted to float8

### DIFF
--- a/scripts-available/CDB_Overviews.sql
+++ b/scripts-available/CDB_Overviews.sql
@@ -15,6 +15,15 @@ AS $$
   END;
 $$ LANGUAGE PLPGSQL IMMUTABLE;
 
+-- Maximum zoom level usable with integer coordinates
+CREATE OR REPLACE FUNCTION _CDB_MaxZoomLevel()
+RETURNS INTEGER
+AS $$
+  BEGIN
+    RETURN 31;
+  END;
+$$ LANGUAGE PLPGSQL IMMUTABLE;
+
 -- Information about tables in a schema.
 -- If the schema name parameter is NULL, then tables from all schemas
 -- that may contain user tables are returned.
@@ -314,7 +323,11 @@ AS $$
     WITH RECURSIVE t(x, y, z, e) AS (
       WITH ext AS (SELECT _cdb_estimated_extent(%6$s) as g),
       base AS (
-        SELECT (-floor(log(2, (greatest(ST_XMax(ext.g)-ST_XMin(ext.g), ST_YMax(ext.g)-ST_YMin(ext.g))/(%4$s*%5$s))::numeric)))::integer z
+        SELECT
+          least(
+           -floor(log(2, (greatest(ST_XMax(ext.g)-ST_XMin(ext.g), ST_YMax(ext.g)-ST_YMin(ext.g))/(%4$s*%5$s))::numeric)),
+           _CDB_MaxOverviewLevel()+1
+          )::integer z
         FROM ext
       ),
       lim AS (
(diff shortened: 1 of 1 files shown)


jgoizueta:
`CDB_ZoomFromScale` returns a maximum zoom level of 23 (for similar reasons to our limiting max. overviews level to 23 too?).

```
    WHEN scaleDenominator <= 100 THEN RETURN 23;
```

So, if we create overviews for level 23 the base table will never be used by Mapnik (images an utfgrid).

Is this a feature or a bug?

* Feature: If a table requires overviews for such high z, we better not try to render it ever (all points have to be very close together); overviews will make sure that rendering is fast.
* Bug: eventually we should be able to see and get info from individual points and not aggregated data

If the bug side wins, we could make the overviews maximum level to be 22, or extend `CDB_ZoomFromScale` to return at least level 24 (but I still don't know what implications this may have).

cc @rafatower @javisantana

javisantana:
the thing is, why we don't use a real math function? https://github.com/mapbox/postgis-vt-util/blob/master/src/Z.sql

jgoizueta:
Regarding using a formula for Z: there's a PR for that: https://github.com/CartoDB/cartodb-postgresql/pull/197

jgoizueta:
I finally found the cause of the floating point arguments error!
It happened when the table had columns named `x` or `y`. In all cases observed these were floating point columns.

The recursive query introduces the names `x` and `y`:

````
    WITH RECURSIVE t(x, y, z, e) AS (
````

Which are used (unqualified) among other places in an inner subquery from the table: 

```
      SELECT x*2 + xx, y*2 + yy, t.z+1, (
        SELECT count(*) FROM %1$s
          WHERE the_geom_webmercator && CDB_XYZ_Extent(t.x*2 + c.xx, t.y*2 + c.yy, t.z+1)
      )
```

And in this case, if the table has and `x`/`y` column the names resolve to that column instead of the recursive query arguments.

Here these names must be qualified as `t.x` and `t.y`  to solve the problem.

**Comment by javisantana** [comment]
good catch!

**Comment by jgoizueta** [comment]
The maximum `CDB_ZoomFromScale` problem being dealt with in #259

**Comment by jgoizueta** [comment]
@rafatower please review

File: scripts-available/CDB_Overviews.sql [hunk 1]
```diff
@@ -1,3 +1,29 @@
+-- Maximum zoom level for which overviews may be created
+CREATE OR REPLACE FUNCTION _CDB_MaxOverviewLevel()
```
rafatower: why not changing the internal representation of coordinates to `bigint` to be able to hold them up to z = 31?

functions such as `CDB_XYZ_Extent` and others can have their `bigint` counterpart. Both can coexist and the `integer` version be marked as "deprecated" and make them issue a warning.

jgoizueta: By using bigints we could raise the maximum overview level from 23 to 29 or 30 (it'd be pointless to support higher levels because using such levels would cause problems elsewhere).
I hadn't think of that approach, basically because I think having up to 23 overview levels should be enough for most cases, but I can do a quick check to see if it a) is easy to implement b) doesn't have a performance impact (the feature density query is already quite slow).

jgoizueta:
I'm keeping the CASE to avoid converting the function to PL/PGSQL

jgoizueta:
Using bigints to avoid the zoom-23 limit seams feasible but also seems to have an impact on performance, since checking that will take some time I'll leave it for later (it's here in this [branch](https://github.com/CartoDB/cartodb-postgresql/compare/bigint_overviews)), as I prefer to deploy the current fixes ASAP.

PR #259, opened 2016-05-27
Use standard formula for CDB_ZoomFromScale

jgoizueta:
This comes from #197 

The maximum zoom, previously limited to 23 is now limited to a higher value, and is returned even
when the scale denominator is 0 (for which the actual level would be infinite) to avoid breaking
some cases where a 0 is used as a dummy scale denominator as here:
https://github.com/CartoDB/Windshaft/blob/2cd43d4b2ca2b3487ede8423f698b282c6e53af4/lib/windshaft/backends/attributes.js#L81

This would be a problem for example for the queries used for overview tables, which will yield no result for a null zoom level, causing that example to fail.

### DIFF
--- a/scripts-available/CDB_ZoomFromScale.sql
+++ b/scripts-available/CDB_ZoomFromScale.sql
@@ -1,30 +1,12 @@
-CREATE OR REPLACE FUNCTION cartodb.CDB_ZoomFromScale(scaleDenominator numeric) RETURNS int AS $$
-BEGIN
+CREATE OR REPLACE FUNCTION cartodb.CDB_ZoomFromScale(scaleDenominator numeric)
+RETURNS int
+LANGUAGE SQL
+IMMUTABLE
+AS $$
+SELECT
   CASE
-    WHEN scaleDenominator > 500000000 THEN RETURN 0;
-    WHEN scaleDenominator <= 500000000 AND scaleDenominator > 200000000 THEN RETURN 1;
-    WHEN scaleDenominator <= 200000000 AND scaleDenominator > 100000000 THEN RETURN 2;
-    WHEN scaleDenominator <= 100000000 AND scaleDenominator > 50000000 THEN RETURN 3;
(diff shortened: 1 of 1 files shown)


jgoizueta:
@rafatower please review

scripts-available/CDB_ZoomFromScale.sql
```diff
@@ -1,30 +1,27 @@
...
-    WHEN scaleDenominator <= 750 AND scaleDenominator > 500 THEN RETURN 20;
-    WHEN scaleDenominator <= 500 AND scaleDenominator > 250 THEN RETURN 21;
-    WHEN scaleDenominator <= 250 AND scaleDenominator > 100 THEN RETURN 22;
-    WHEN scaleDenominator <= 100 THEN RETURN 23;
-  END CASE;
-END
-$$ LANGUAGE plpgsql IMMUTABLE;
+    -- Don't bother if the scale is larger than ~zoom level 0
+    WHEN scaleDenominator > 600000000 OR scaleDenominator = 0 THEN NULL
+    WHEN scaleDenominator = 0 THEN _CDB_MaxSupportedZoom()
```
rafatower: There's a contradiction regarding how `scaleDenominator = 0` case is treated. Shall it return `NULL` or `_CDB_MaxSupportedZoom()` ?

this `WHEN` is never going to be executed AFAIK.

rafatower: I'd use an if-else instead

jgoizueta: oops I failed to remove the `OR scaleDenominator = 0`, thanks

jgoizueta: `IF ELSE` should be better here, I was lazy and kept the existing structure

jgoizueta: Actually, the reason for using CASE originally seems to be this is written as SQL, not PL/PGSQL.

jgoizueta:
I'm keeping the CASE to avoid converting the function to PL/PGSQL

rafatower:
:+1:

PR #262, opened 2016-06-07
Exclude overviews from user data size

### LINKED ISSUE #261: Should overviews tables size count for the user quota?
Currently for raster table we count the base table size (fully) but not its overviews.
For regular tables (and its vector overviews) we count half of its size (to account for the overhead of having the webmercator geometry).

If we apply the same criterion to vector overviews as we do for raster overviews we should not count it's size... or we could decide to count a fraction of its size...

jgoizueta:
Fixes #261


scripts-available/CDB_Quota.sql
```diff
@@ -35,7 +35,9 @@ BEGIN
   table_cat AS (
     SELECT
       table_name,
-      EXISTS(select * from raster_tables where o_table_name = table_name) AS is_overview,
+      EXISTS(select * from raster_tables where o_table_name = table_name)
+      OR table_name SIMILAR TO _CDB_OverviewTableDiscriminator() || '[\w\d]*'
+      AS is_overview,
```
rafatower: a couple of parenthesis or indentation can improve legibility

scripts-available/CDB_OverviewsSupport.sql
```diff
@@ -0,0 +1,173 @@
+-- Auxiliary overviews FUNCTIONS
```
rafatower: Renaming a file for a bugfix really obscures the review. Have you actually modified anything in this file

jgoizueta: mmm, I stashed all my messy commits together and now the actual fix isn't clear at all.
But it's not a rename, it's just some functions from Overviews that are or could be used from other files have been extracted to a separate file.

rafatower: I see, my mistake. Thanks for the clarification :smile:

scripts-available/CDB_OverviewsSupport.sql
```diff
@@ -0,0 +1,173 @@
...
+    RETURN 31;
+  END;
+$$ LANGUAGE PLPGSQL IMMUTABLE;
+
+-- Information about tables in a schema.
+-- If the schema name parameter is NULL, then tables from all schemas
+-- that may contain user tables are returned.
+-- For each table, the regclass, schema name and table name are returned.
+-- Scope: private.
+CREATE OR REPLACE FUNCTION _CDB_UserTablesInSchema(schema_name text DEFAULT NULL)
```
rochoa: Sorry to stick my nose into this one, but shouldn't this function sleep under `CDB_UserTables.sql`?

And also other related helper functions: `_cdb_split_table_name`, `_cdb_schema_name`?

jgoizueta: That makes sense. These internal functions were written to support the `Overviews.sql` and honestly, I didn't pay much attention to where to put them since they're internal.

But we're using 'internal' functions accross source files, so we may as well put some order into it, I guess.

rafatower: IMHO only if there are plans to use them outside this file. Otherwise I'd keep them as "private" and close to the place where they are used (that is, here, right?)

rochoa: Then the problem for me is ... For the very same reason why aren't they placed in `CDB_Overviews.sql`?

PR #267, opened 2016-07-01
Add some extra fields to cdb_analysis_catalog table

### LINKED ISSUE #265: CDB_AnalysisCatalog should allow to store error messages for nodes
So it's easier to share error messages from different applications operating it.

rochoa:
Track user, error_message for failures, and last entity modifying the node.

Closes #265.


rochoa:
cc @luisbosque @zenitraM @rafatower 

As we don't have upgrade/downgrade paths in this extension, do you know a better way to add columns to an existing table without dropping the table?

zenitraM:
Could work - only thing I'm worried about is the NOTICEs we will get on every extension upgrade from now on.

rochoa:
Let me try to fix that :-).

rafatower:
I don't think it can be done much better, so OK.

PR #270, opened 2016-07-05
Exclude null geometries to create overviews

### LINKED ISSUE #269: Exclude NULL geometries when creating Overviews
The presence of (many) NULL geometries can make overview creation take a long time (more than the current 1h timeout used in CartoDB).

Simply filtering out NULL geometries here https://github.com/CartoDB/cartodb-postgresql/blob/415d96082e5b6cef72299fef89d4f170d911aa02/scripts-available/CDB_Overviews.sql#L689
(`WHERE the_geom IS NOT NULL`) avoids the problem.

An inconvenience of doing this is that if we decide to create overviews based on the estimate of the row count of a table, we may end up with overviews with few records that shouldn't have been created in the first place. (and counting the number of rows with non-null geometry can be costly).

Note that `_CDB_Feature_Density` is already ignoring records with null geometries, since it performs spatial filtering.

jgoizueta:
Fixes #269

### DIFF
--- a/scripts-available/CDB_Overviews.sql
+++ b/scripts-available/CDB_Overviews.sql
@@ -687,6 +687,7 @@ AS $$
              Floor(ST_Y(f.the_geom_webmercator)/%2$s)::int AS gy,
              MIN(cartodb_id) AS cartodb_id
           FROM %1$s f
+          WHERE f.the_geom_webmercator IS NOT NULL
           GROUP BY gx, gy
          )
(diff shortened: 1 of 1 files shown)


rafatower:
Do you feel confident as to release this fix in the next deploy? just found this looking for another PR :)

jgoizueta:
Yes, I had in mind including this in the nest release. It was tested.

PR #274, opened 2016-08-11
Adding a new column to have the name of the cached analysis in case there is one

### LINKED ISSUE #273: Add new column to cdb_analysis_catalog
A new column for cdb_analysis_catalog with the name of the possible cached table associated with the analysis would be helpfulu for cleaning tasks.

azamorano:
Fixes #273 

@rochoa please review!


rochoa:
Although this is enough, I would add the column also to the table creation to document what is the target of the column :-).

Issue #279, opened 2016-10-11
Function to check analysis tables limits


jgoizueta:
This is to support https://github.com/CartoDB/camshaft/issues/206

We need a function `CDB_CheckAnalysisQuota(table_name TEXT)` that raises an exception if the limit for Camshaft cached analysis tables has been exceeded.

This will be called after each cache table is populated, passing the name of the table as a parameter. For our initial strategy (of just limiting the total analysis table space per user as a fraction of his/her quota) we don't really need to use the table name parameter, but I'd leave it there so we can try different approaches later.

PR #280, opened 2016-10-11
Implement CDB_CheckAnalysisQuota

jgoizueta:
see #279

### DIFF
--- a/scripts-available/CDB_AnalysisCheck.sql
+++ b/scripts-available/CDB_AnalysisCheck.sql
@@ -1,0 +1,50 @@
+-- Get the factor (fraction of the quota) for Camshaft cached analysis tables
+CREATE OR REPLACE FUNCTION _CDB_AnalysisQuotaFactor()
+RETURNS float8 AS
+$$
+DECLARE
+  factor float8;
+BEGIN
+  -- We use a floating point cdb_conf parameter
+  SELECT cdb_conf_getconf::text::float8 INTO factor FROM CDB_Conf_GetConf('analysis_cache_factor');
+  -- With a default value
+  IF factor IS NULL THEN
+    factor := 0.2;
+  END IF;
+  RETURN factor;
+END;
+$$
+LANGUAGE 'plpgsql' STABLE;
+
+-- This checks the space used up by Camshaft cached analysis tables.
+-- An exception will be raised if the limits are exceeded.
+-- The name of an analysis table is passed; this, in addition to the
+-- db role that executes this function is used to determined which
+-- analysis tables will be considered.
+CREATE OR REPLACE FUNCTION CDB_CheckAnalysisQuota(table_name TEXT)
+RETURNS void AS
+$$
+DECLARE
+  schema_name TEXT;
+  user_name TEXT;
+  qmax int8;
+  cache_size float8;
+BEGIN
+  -- We rely on the search_path to determine the user's schema and
+  -- check for all analysis tables in that schema.
+  -- An alternative would be to use cdb_analysis_catalog to
+  -- select analysis tables (cache_tables) from the same user, analysis or node.
+  -- For example:
+  --   SELECT unnest(cache_tables) FROM cdb_analysis_catalog
+  --     WHERE username IN (SELECT username FROM cdb_analysis_catalog
+  --       WHERE table_name::regclass = ANY (cache_tables));
+  -- At the moment we're not using the provided table_name.
+
+  SELECT current_schema() INTO schema_name;
+  EXECUTE FORMAT('SELECT %I._CDB_UserQuotaInBytes();', schema_name) INTO qmax;
+  IF qmax*_CDB_AnalysisQuotaFactor() < _CDB_AnalysisDataSize(schema_name) THEN
+    -- The limit is defined by a factor applied to the total space quota for the user
+    RAISE EXCEPTION 'Analysis cache space limits exceeded';
+  END IF;
+END;
+$$ LANGUAGE PLPGSQL;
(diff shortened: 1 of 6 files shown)


jgoizueta:
@ethervoid @rafatower please CR

jgoizueta:
This will be used from [here](https://github.com/CartoDB/camshaft/pull/207)

scripts-available/CDB_AnalysisCheck.sql
```diff
@@ -0,0 +1,62 @@
+-- Read configuration parameter analysis_quota_factor, making it
+-- accessible to regular users (which don't have access to cdb_conf)
+CREATE OR REPLACE FUNCTION _CDB_GetConfAnalysisQuotaFactor()
+RETURNS float8 AS
+$$
+BEGIN
+  RETURN CDB_Conf_GetConf('analysis_quota_factor')::text::float8;
```
rafatower: mind that these config entries need to be  propagated to all user's DB's. I guess you did because of the default value below.

scripts-available/CDB_AnalysisCheck.sql
```diff
@@ -0,0 +1,62 @@
...
+
+-- Get the factor (fraction of the quota) for Camshaft cached analysis tables
+CREATE OR REPLACE FUNCTION _CDB_AnalysisQuotaFactor()
+RETURNS float8 AS
+$$
+DECLARE
+  factor float8;
+BEGIN
+  -- We use a floating point cdb_conf parameter
+  factor := _CDB_GetConfAnalysisQuotaFactor();
```
rafatower: does it return null if not present? very nice it has a default value

scripts-available/CDB_AnalysisSupport.sql
```diff
@@ -0,0 +1,55 @@
...
+CREATE OR REPLACE FUNCTION _CDB_NonAnalysisTablesInSchema(schema_name text DEFAULT NULL)
+RETURNS TABLE(table_regclass REGCLASS, schema_name TEXT, table_name TEXT)
+AS $$
+  SELECT * FROM _CDB_UserTablesInSchema(schema_name) WHERE Not _CDB_IsAnalysisTableName(table_name);
+$$ LANGUAGE 'sql';
+
+-- Total spaced used up by Camshaft cached analysis tables in the given schema.
+-- Scope: private.
+CREATE OR REPLACE FUNCTION _CDB_AnalysisDataSize(schema_name TEXT DEFAULT NULL)
+RETURNS bigint AS
```
ethervoid: Why return bigint when you're returning an int8?

scripts-available/CDB_AnalysisSupport.sql
```diff
@@ -0,0 +1,55 @@
...
+$$
+DECLARE
+  total_size INT8;
+BEGIN
+  WITH analysis_tables AS (
+    SELECT t.schema_name, t.table_name FROM _CDB_AnalysisTablesInSchema(schema_name) t
+  )
+  SELECT COALESCE(INT8(SUM(_CDB_total_relation_size(analysis_tables.schema_name, analysis_tables.table_name))))::int8
```
ethervoid: You're adding up all the analysis tables for that user but AFAIK we want to check every analysis table independently. I know that one analysis could have multiple tables, maybe we should think how to group them.

If this is the intended behaviour, please ignore this comment

scripts-available/CDB_AnalysisSupport.sql
```diff
@@ -0,0 +1,55 @@
...
+AS $$
+  SELECT * FROM _CDB_UserTablesInSchema(schema_name) WHERE _CDB_IsAnalysisTableName(table_name);
+$$ LANGUAGE 'sql';
+
+-- This function returns a relation user tables excluding analysis tables
+-- If the schema name parameter is NULL, then tables from all schemas
+-- that may contain user tables are returned.
+-- For each table, the regclass, schema name and table name are returned.
+-- Scope: private.
+CREATE OR REPLACE FUNCTION _CDB_NonAnalysisTablesInSchema(schema_name text DEFAULT NULL)
```
ethervoid: This function looks unused

jgoizueta: Yeah, I got carried away and put that thinking we'll need it to fix #277 but it is still unused.

scripts-available/CDB_AnalysisSupport.sql
jgoizueta: At the moment we're just limiting the total cached space for the user, so this is intended.

It could make sense as you mention to limit per analysis and could be done using `cdb_analysis_catalog`, but we'll consider it later. There's a comment about that here: https://github.com/CartoDB/cartodb-postgresql/pull/280/files#diff-eece45bf20ce8953f780763c3a0916d0R47

scripts-available/CDB_AnalysisSupport.sql
jgoizueta: Oops, they're supposed the be the same (aliases), and I inadvertently used both names. I'll fix it, but which one do you like better? `bigint`or  `int8`?

scripts-available/CDB_AnalysisCheck.sql
jgoizueta: Yes, I'm aware. We're using this as an easy way to be able to modify the default, for tests or so... We may think of something better in the future.

scripts-available/CDB_AnalysisSupport.sql
ethervoid: I think bigint is better naming

rafatower: changed

Issue #157, opened 2015-09-18
Test failing


Tom-Tom:
In https://github.com/CartoDB/cartodb/issues/5421

> If I run installcheck after installing postgis, I get this error:
> `> Executing query 'SELECT count(*) FROM cdb_testmember_1.foo;' as cdb_testmember_2 ERROR: permission denied for schema cdb_testmember_1 LINE 1: SELECT count(*) FROM cdb_testmember_1.foo; ^ FAILED TO EXECUTE QUERY: SELECT count(*) FROM cdb_testmember_1.foo;`
> 
> And this fun one, which might be expected: `ERROR: relation "unexistent_example" does not exist LINE 1: SELECT CDB_TableMetadataTouch('unexistent_example'); ^ FAILED TO EXECUTE QUERY: SELECT CDB_TableMetadataTouch('unexistent_example');`
> 
> And at the end: The following tests are failing: `test_member_2_cannot_read_without_permission`

I have the same problem.

odessky:
Confirmed. I have same error on Ubuntu 12.04 amd64.

Algunenano:
Should be ok now

PR #283, opened 2016-10-17
Fix tests: race condition with publicuser #157

rafatower:
There's a test race between the `test/extension/test.sh` and the `test/organization/test.sh`.

Particularly the fdw test count on the `publicuser` being present but then the organization tests count on just the opposite. It is hard to relate the cause with the effect (see the traces posted in the ticket) but believe me, this fixes (that instance of) the problem.

`publicuser` is meant to be a system requirement, so I guess we can get around this and other test races just by assuming it is there or creating it on demand, pretty much like in this PR. The `DROP ROLE` should be optional.

### DIFF
--- a/test/extension/test.sh
+++ b/test/extension/test.sh
@@ -227,6 +227,7 @@ function tear_down() {
     sql 'DROP ROLE cdb_testmember_2;'
 
     tear_down_database
+    DATABASE=postgres sql postgres 'DROP ROLE IF EXISTS publicuser';
 }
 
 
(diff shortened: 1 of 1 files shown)


rafatower:
This fixes #157 

@jgoizueta can you please review?

cc/ @juanignaciosl . I'll work on a more robust solution as mentioned above. Also I'm playing with a `Dockerfile` that I think we can use both for travis and local development/testing (many people have trouble with the setup). So far I also found an incompatibility with 9.5 easy to solve, but don't want to break CI until I get everything working.

test/extension/test.sh
```diff
@@ -227,6 +227,7 @@ function tear_down() {
     sql 'DROP ROLE cdb_testmember_2;'
 
     tear_down_database
+    DATABASE=postgres sql postgres 'DROP ROLE IF EXISTS publicuser';
```
jgoizueta: Will the tests fail if there are dependencies on `publicuser`? (e.g. for some reason in my local env. other databases are using that role).

rafatower: it can fail, but as stated in the readme:
https://github.com/CartoDB/cartodb-postgresql#test-installation
> NOTE: the tests need to run against a clean postgres instance, if you have some roles already created test will likely fail due publicuser not being dropped.

rafatower:
I'll defer this. I think the correct approach is to remove all the `DROP ROLE publicuser` present in tests.

PR #285, opened 2016-10-20
Move "username" column to the last position on analysis_catalog - fixes #276

### LINKED ISSUE #276: Make cdb_analysis_catalog content be dumpable and restoreable
We only care about schema on pg_dump/restore.

cc @zenitraM

zenitraM:
Due to the way it was first implemented, the "username" column would be on a different position depending on if it was an extension upgrade or a fresh install.

The cause for this was the varying column order we introduced on one of the migrations on 064b26ccd327bf079a9d25f9ed2fe79c2c742bae, in particular, the username field.

As we added the username on the last column for new extension installs but on a different one on upgrades, this caused the column order not be always the same. (column_index 1 for newly-installed after 0.17.0 or 10 for upgraded from a previous version).

pg_dump does not include the extension source (so this table's schema is not included on the dump) but does include its contents, using a COPY without column names by default and so failing to restore extensions that went through an upgrade due to the order difference.

After talking with @rochoa, to be consistent on the way forward and not need to run any "transient" fixes, and due to the lack of a way to reorder columns on Postgres, we have decided to move the username to the last position of the table on both cases from this version on.

cc @rochoa @rafatower @luisbosque

### DIFF
--- a/scripts-available/CDB_AnalysisCatalog.sql
+++ b/scripts-available/CDB_AnalysisCatalog.sql
@@ -1,8 +1,6 @@
 -- Table to register analysis nodes from https://github.com/cartodb/camshaft
 CREATE TABLE IF NOT EXISTS
 cartodb.cdb_analysis_catalog (
-    -- useful for multi account deployments
-    username text,
     -- md5 hex hash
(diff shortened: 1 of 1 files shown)


rafatower:
ack. :+1:

rochoa:
🎉  👍

PR #288, opened 2016-11-04
Fix migration script generation

jgoizueta:
In the 0.18.3 release the script to migrate from 0.18.2 was missing.

This will generate a new version 0.18.4 that when installed will generate scripts to migrate from all old versions to it, so it will be possible to migrate existing users to 0.18.4 (but not to 0.18.3)


ethervoid:
lgtm

Issue #286, opened 2016-10-28
Overview strategy with point samples


jaakla:
Instead of putting clusters to pixel grid centres use one of the points as a sample. This gives nicer look if you zoom in map quickly using vector data (in mobile-sdk) - with default strategy you see regular grids of points as pixels until more detailed zoom is loaded.

jgoizueta:
The strategy currently used for automatically created overviews on imports (grid centre) was chosen for browser tiled visualisation mimicking the approach of torque, and assuming the grid size would be smaller than a pixel in normal visualisation.

When creating overviews manually (i.e. calling `CDB_CreateOverviews`) we can provide a different clustering strategy (in the form of a PostgreSQL function).

I think we could some additional strategies to the cartodb extension so they're available without having to define new functions, and we could add also an optional parameter to the Import API to be able to choose the overviews strategy (and grid size).

Example: if we add a function named `_CDB_GridClusterSample_Reduce_Strategy` to the extension (which could implement a clustering strategy that places the aggregated cluster points at the location of some of the original points) we can use it with (choosing also a grid size of 8 pixels):

```
select CDB_CreateOverviewsWithToleranceInPixels('table', 8, '_CDB_Feature_Density_Ref_Z_Strategy(REGCLASS,FLOAT8)'::regprocedure, 'CDB_GridClusterSample_Reduce_Strategy(REGCLASS,INTEGER,INTEGER,FLOAT8)'::regprocedure);
```

Here are a couple of functions that implement different clustering strategies: https://github.com/CartoDB/cartodb-postgresql/commit/5992304b475232a3f95b9a628b9f83e82eae1d43

jaakla:
Maybe even the default strategy could be replaced, as for rasters there is probably no visible difference if we put the aggregated point to the center of pixel or to some other point within same pixel? For vector tiles (which is the future anyway) Sample-based strategy would be much better, especially for quick zoom in and overzoom use cases. I have used 8 and 16-pixel tolerances for vector tile size control with good results.

jgoizueta:
The current default was carefully adjusted to act like torque clustering to a 1-pixel grid, and in particular to avoid visual artifacts (at pixel level) due to the use of TWKB (see #240).

[...]

PR #290, opened 2016-11-29
Add point overview strategies

jgoizueta:
See #286


jgoizueta:
@rtorre please review; we're only adding private functions it shouldn't be a problem to release this now. cc @ethervoid

rafatower:
LGTM

Issue #295, opened 2017-03-15
Create function CDB_EstimateRowCount(query text)


rafatower:
### Description

Add a new function to `cartodb` extension to be called `CDB_EstimateRowCount(query text)`.

The purpose of that function is, given a query, to return a row count estimate that is suitable for quota usage estimation.

Example of its usage:
```sql
SELECT CDB_EstimateRowCount($$
  UPDATE addresses SET the_geom = cdb_geocode_street_point(addr, city, state, 'US');
$$) AS row_count;

 row_count 
-----------
         5
(1 row)
```

What the function should do:
1. Get the list of tables affected by the query
1. Make sure stats exist for all of them, otherwise execute `ANALYZE` to get them.
1. Use `EXPLAIN` to get a row count estimate of the query.

### Implementation notes
* The list of affected tables can be get by using [CDB_QueryTablesText](https://github.com/CartoDB/cartodb-postgresql/blob/master/scripts-available/CDB_QueryTables.sql)
* The view `pg_stat_user_tables` is a convenient place to look for stats.
* There's still a corner case with *fresh empty tables* that we don't aim to solve here. See [this](https://github.com/postgres/postgres/blob/18dc2aee5f303447bef48dee596a664d90f6939a/src/backend/optimizer/util/plancat.c#L915-L951)

jgoizueta:
For reference, see this JS implementation: https://github.com/CartoDB/camshaft/blob/054fb088ffdae8f3e89c455ebff12dd323ca6693/lib/limits/estimates.js#L27-L38

jgoizueta:
PR here: https://github.com/CartoDB/cartodb-postgresql/pull/298

jgoizueta:
Acceptance could be checked in staging, installing the branch and updating the cartodb-extension for some user.

Things to check:
* For a newly created table with some rows inserted the estimate `CDB_EstimateRowCount('select * from table')` should be accurate. Also: for a fresh table with no stats, after calling `CDB_EstimateRowCount` stats should be present.
* For a newly created empty table an estimation (albeit not accurate) should be obtained (typically 850 rows).
* For older populated tables some sensible estimation should be obtained
* It should return a reasonable estimate for filtered queries
* It should work for org user tables. Extra bonus: check with org username containing a hyphen.
* It should fail with an error with an invalid query

jgoizueta:
Acceptance was successful in staging; moving on to deploy in production

jgoizueta:
[...]

PR #298, opened 2017-04-07
Add CDB_EstimateRowCount function

jgoizueta:
See #295


jgoizueta:
What could be the reason for the failing test?:

```
*** /home/travis/build/CartoDB/cartodb-postgresql/expected/test/CDB_QueryTablesTest.out	2017-04-07 14:54:03.627042459 +0000
--- /home/travis/build/CartoDB/cartodb-postgresql/results/test/CDB_QueryTablesTest.out	2017-04-07 14:55:04.715253946 +0000
***************
*** 5,16 ****
  SELECT a.oid, b.oid FROM pg_class a, pg_class b|{pg_catalog.pg_class}
  SELECT 1 as col1; select 2 as col2|{}
  WARNING:  CDB_QueryTables cannot explain query: select 1 from nonexistant (42P01: relation "nonexistant" does not exist)
- CONTEXT:  PL/pgSQL function cdb_querytables(text) line 3 at RETURN
  ERROR:  relation "nonexistant" does not exist
  CONTEXT:  PL/pgSQL function cdb_querytables(text) line 3 at RETURN
  begin; select * from pg_class; commit;|{pg_catalog.pg_class}
  WARNING:  CDB_QueryTables cannot explain query: select * from test (42P01: relation "test" does not exist)
- CONTEXT:  PL/pgSQL function cdb_querytables(text) line 3 at RETURN
  ERROR:  relation "test" does not exist
  CONTEXT:  PL/pgSQL function cdb_querytables(text) line 3 at RETURN
  WITH a AS (select * from pg_class) select * from a|{pg_catalog.pg_class}
--- 5,14 ----
```

Maybe the message level is changing depending on the order the tests are executed?

rafatower:
There are two different concepts when logging stuff in PG: "verbosity" and "level". When it is `verbose`, it spits the `CONTEXT`, kind of stack trace. When it is set to `terse`, it doesn't give you that context. Aside the level determines whether to log something or not at all (ie: debug, info, warning, error).

HTH

test/CDB_EstimateRowCountTest.sql
```diff
@@ -0,0 +1,10 @@
+SET client_min_messages TO error;
+\set VERBOSITY terse
```
rafatower: this looks like the cause of your "test races"

scripts-available/CDB_EstimateRowCount.sql
```diff
@@ -0,0 +1,31 @@
...
+  has_stats BOOLEAN;
+BEGIN
+  SELECT EXISTS (
+    SELECT * FROM pg_catalog.pg_statistic WHERE starelid = reloid
+  ) INTO has_stats;
+  IF NOT has_stats THEN
+    EXECUTE Format('ANALYZE %s;', reloid);
+  END IF;
+END
+$$ LANGUAGE 'plpgsql' VOLATILE STRICT SECURITY DEFINER;
```
rafatower: Just curious: did you find any need to actually have `SECURITY DEFINER` here?

scripts-available/CDB_EstimateRowCount.sql
```diff
@@ -0,0 +1,31 @@
...
+RETURNS VOID
+AS $$
+DECLARE
+  has_stats BOOLEAN;
+BEGIN
+  SELECT EXISTS (
+    SELECT * FROM pg_catalog.pg_statistic WHERE starelid = reloid
+  ) INTO has_stats;
+  IF NOT has_stats THEN
+    EXECUTE Format('ANALYZE %s;', reloid);
```
rafatower: are you counting on automatic casting of `text` to `reloid` ? does it work well with org user's tables?

jgoizueta: Since `reloid` is a REGCLASS here, format %s will properly include the schema and doublequote it if necessary.

Then, I'm relying on `CDB_QueryTablesText` returning schema names as necessary, which seems to be the case. (the results of that function are casted to REGCLASS to be passed here).

So, it seems to work well with org user's, even when the username requires double quotes.

scripts-available/CDB_EstimateRowCount.sql
jgoizueta: It's needed to query `pg_catalog.pg_statistic`:

> pg_statistic should not be readable by the public, since even statistical information about a table's contents might be considered sensitive. (Example: minimum and maximum values of a salary column might be quite interesting.) pg_stats is a publicly readable view on pg_statistic that only exposes information about those tables that are readable by the current user.

I considered using the `pg_stats` view instead to get rid of the SECURITY DEFINER but I didn't in the end (because of my laziness to find out if using it would make for any difference in behaviour).

rafatower: oh, I see. thanks for the clarification

scripts-available/CDB_EstimateRowCount.sql
rafatower: :+1:

PR #302, opened 2017-06-02
Regenerate overviews deadlock

ethervoid:
Instead of make a DROP/CREATE always, in case we have existing
overviews we have to make DELETE/INSERT because DDL operations require
AccessExclusiveLock which is not compatible with AccessShareLock that
is used by the queries used to gather data for the tiler.

This incompatibility above mentioned leads to have deadlocks


scripts-available/CDB_Overviews.sql
```diff
@@ -904,14 +808,20 @@ AS $$
       columns := columns || ', n AS _feature_count';
     END IF;
 
-    EXECUTE Format('DROP TABLE IF EXISTS %I.%I CASCADE;', schema_name, overview_rel);
+    overview_table_name := Format('%I.%I', schema_name, overview_rel);
```
ethervoid: We should extract this somehow and make the *_Reduce_Strategy functions return the data not make the tables, populate them and so on. I didn't want to change this becuase is a huge change...

scripts-available/CDB_Overviews.sql
```diff
@@ -671,131 +681,22 @@ AS $$
...
-      offset_y := Format('%s', offset_m);
+    overview_table_name := Format('%I.%I', schema_name, overview_rel);
+    IF has_overview_created THEN
+      RAISE INFO 'Deleting and inserting because % has overviews', overview_table_name;
+      EXECUTE Format('DELETE FROM %s;', overview_table_name);
+      creation_clause := Format('INSERT INTO %s', overview_table_name);
     ELSE
-      offset_x := Format('%2$s/2 - MOD((%1$s)::numeric, (%2$s)::numeric)::float8', cell_x, pixel_m);
-      offset_y := Format('%2$s/2 - MOD((%1$s)::numeric, (%2$s)::numeric)::float8', cell_y, pixel_m);
+      RAISE INFO 'Creating a new table overview %', overview_table_name;
```
ethervoid: Maybe NOTICE?

scripts-available/CDB_Overviews.sql
```diff
@@ -284,21 +292,21 @@ AS $$
...
             ARRAY[
               (SELECT CDB_RandomTids(''%2$s'', %3$s))
             ]
           );
-      ', overview_rel, reloid, num_samples, schema_name);
+      ', creation_clause, reloid, num_samples);
     END IF;
 
-    RETURN Format('%I.%I', schema_name, overview_rel)::regclass;
+    RETURN Format('%s', overview_table_name)::regclass;
```
jgoizueta: why has the schema_name been removed?

scripts-available/CDB_Overviews.sql
jgoizueta: Agreed, it should be refactored. I also agree we can leave it out for a latter ticket.

scripts-available/CDB_Overviews.sql
ethervoid: I created a a variable which is compounded by schema_name and overview_rel. You can see it [here](https://github.com/CartoDB/cartodb-postgresql/pull/302/files/09309cc100b99e14513da13ae4de9dc242301723#diff-f3f06dde01c50fc3d5586076be12f211R280)

PR #306, opened 2017-06-29
305 fix unique identifiers

rafatower:
@jgoizueta can you please take a quick and gentle look at this? unless you are strongly opposed I'll merge and modify the release files in master, then deploy


rafatower:
PS: fixing tests took me a long while... if we're to work more on this, we should think about changing them to use either the old/new framework or pgtap or something...

Issue #266, opened 2016-06-29
CDB_QuantileBins producing null stops


rochoa:
Example:

```
https://documentation.cartodb.com/api/v1/sql?q=SELECT%20CDB_QuantileBins%28ARRAY%5B%5D::numeric[],%205%29
```

andy-esch:
This is fixed in https://github.com/CartoDB/cartodb-postgresql/pull/316

PR #307, opened 2017-08-02
Remove nulls

andy-esch:
Null values should not be considered when calculating the quantile breaks.

Key line is this one: https://github.com/CartoDB/cartodb-postgresql/blob/40f902012ae03b1b40844a06d1f694f6ffed55d2/scripts-available/CDB_QuantileBins.sql#L22

```sql
-- sort our values, remove null values
SELECT array_agg(e) INTO in_array
 FROM (SELECT unnest(in_array) e
       ORDER BY e ASC) AS x
WHERE e IS NOT NULL;
```

cc @rochoa @stuartlynn


andy-esch:
Looks like the test work for the new `CDB_QuantileBins_test`, but fail for Jenks, which I did not touch.

jgoizueta:
While you're at it... shouldn't NaNs be also filtered out?

On the other hand, infinities can arguably be included, since they have a numerical well defined order.

andy-esch:
@jgoizueta good idea about `NaN`s, I forgot about them already :) I'll change the filter to this:

```sql
...
WHERE col != 'NaN' and 
      col is not null
```

BTW, I'm getting strange behavior of `Infinities` between numeric and float types. E.g., If I cast a column which has `Infinity`s, the column which has the `Infinity` disappears:

infs

Through the SQL API it appears as normal: http://eschbacher.cartodb.com/api/v2/sql?q=SELECT%20val,%20val::float%20as%20v2%20FROM%20eschbacher.inf_nan_values

Is there a difference between numeric and float for the handling of NaNs and Infs that I'm missing?

jgoizueta:
Float columns can have NaN and Inifinite values, but Numeric function can't have Inifinite values (the do support NaN, though). The behavioiur is weird, though.

andy-esch:
Sorry, I made a mistake about Infinity casting behavior. SQL API has same behavior. I updated my comment above.

@stuartlynn and I talked through use of Infinity use in quantile bins. It seems like there's not a standard practice in other software packages, but most of the time infinities are replaced with NaNs in a data prep step. On top of that, since `numeric` type doesn't support `infinity` my runs that include these values results in no rows returned or errors, like this:

```SQL
WITH data_infs AS (
    SELECT array_agg((CASE WHEN x %25 2 != 0 THEN x::float ELSE 'Infinity' END)::numeric)::numeric[] s
      FROM generate_series(1,100) x
     WHERE x %25 5 != 0 AND x %25 7 != 0
    )
SELECT unnest(CDB_QuantileBins(s, 7)) FROM data_infs
```

SQL API call: http://eschbacher.cartodb.com/api/v2/sql?q=WITH%20data_infs%20AS%20(%20SELECT%20array_agg((CASE%20WHEN%20x%20%25%202%20!=%200%20THEN%20x::float%20ELSE%20%27Infinity%27%20END)::numeric)::numeric[]%20s%20FROM%20generate_series(1,100)%20x%20WHERE%20x%20%25%205%20!=%200%20AND%20x%20%25%207%20!=%200%20)%20SELECT%20unnest(CDB_QuantileBins(s,%207))%20FROM%20data_infs

So given all of this I'm going to remove infinities from the bin calculation.

andy-esch:
@jgoizueta, is there anything I can do to help move this along?

jgoizueta:
I'm assigning this to @rochoa for triage-ing

rochoa:
> Looks like the test work for the new `CDB_QuantileBins_test`, but fail for Jenks, which I did not touch.

@andy-esch, it looks like `CDB_JenksBins` depends on `CDB_QuantileBins`:

https://github.com/CartoDB/cartodb-postgresql/blob/5d43faecaf5659cbc32f3e9468e10e029ea47cd7/scripts-available/CDB_JenksBins.sql#L49-L50

That might explain why the tests started to fail. Can you check what is going on?

andy-esch:
Ah, that makes sense. Yep, I'll take a look

andy-esch:
Ok, it makes sense that that test would fail because it pulls in null values that previously would not have been filtered out.

andy-esch:
@rochoa, which of the options above sound like the best route for moving forward?

rochoa:
What is the goal for this PR? Do we want to just "Remove nulls" or it should be about "_Fixing CDB_QuantileBins with nulls_"?

I would do two separate PR.

andy-esch:
It's goal is to only consider numerical data when calculating the bins so that bin edges do not include null (like `1, 4, 10, null`), which leads to erroneous classification of data on maps.

andy-esch:
Related #266

Issue #318, opened 2018-01-18
Check how we manage PK sequences in cartodbfy


rafatower:
As per @pramsey it seems like we're creating sequences for cartodbfy'ed tables without taking into account ownership. If that is the case, the sequences will be left all over the place when deleting tables.

We should fix it for the sake of housekeeping. Mind that a large number of dangling objects in the catalog may impact performance.

pramsey:
I'm wrong, we do set ownership (though maybe we once didn't? ogr2ogr didn't? anyways I had some orphan sequences, but in general the tables in my account *do* set ownership. Also, though there was a bug, it was really all ogr2ogr's fault. There's one tweak we could make, I'll do a little PR for discussion, it's cosmetic.

Algunenano:
Is this still necessary? Seems that @pramsey confirmed this isn't an issue but I could have a look if required.

rafatower:
Not for RT but to take into account when adding the COPY to GDAL, to take advantage of it to clean up accesses to catalog and metadata that can be avoided in the driver.

pramsey:
Let's shut this one down, I think it happened, but perhaps related to 3rd party tools, or in the distant past.

PR #319, opened 2018-01-18
Sequence Rename to avoid Collision

pramsey:
This doesn't really fix anything on our side, since the existing code already avoids sequence name collisions, but it addresses the issue in #318. This would just allow older versions of ogr2ogr to avoid the bug fixed in [7203](https://trac.osgeo.org/gdal/ticket/7203) where ogr2ogr fixates on having a serial sequence named `<relname>_cartodb_id_seq`. What exposed the error was

* running an `ogr2ogr` data upload (works)
* running it again (fails with errors about missing sequence)

It happens because after the first `ogr2ogr` upload, the table is cartodb'fyed, and the sequence name changes when cartodbfy does the copy step, so it ends up with a unique sequence name that is **not** the  `<relname>_cartodb_id_seq` form `ogr2ogr` expects when running in `-append` mode, which seems to be the default for a not-first upload.

### DIFF
--- a/scripts-available/CDB_CartodbfyTable.sql
+++ b/scripts-available/CDB_CartodbfyTable.sql
@@ -792,7 +792,6 @@ DECLARE
   relname TEXT;
   relschema TEXT;
   relseq TEXT;
-  relseqtmp TEXT;
 
   destoid REGCLASS;
   destname TEXT;
@@ -915,7 +914,7 @@ BEGIN
   -- Put the primary key sequence in the right schema
   -- If the new table is not moving, better ensure the sequence name
   -- is unique
-  destseq := cartodb._CDB_Unique_Identifier(NULL, destseq, destschema);
+  destseq := cartodb._CDB_Unique_Identifier(NULL, relname, '_' || const.pkey || '_seq', destschema);
   destseq := Format('%I.%I', destschema, destseq);
   PERFORM _CDB_SQL(Format('CREATE SEQUENCE %s', destseq), '_CDB_Rewrite_Table');
 
--- a/scripts-available/CDB_CartodbfyTable.sql
+++ b/scripts-available/CDB_CartodbfyTable.sql
@@ -791,6 +791,8 @@ DECLARE
 
   relname TEXT;
   relschema TEXT;
+  relseq TEXT;
+  relseqtmp TEXT;
 
   destoid REGCLASS;
   destname TEXT;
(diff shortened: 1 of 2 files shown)


scripts-available/CDB_CartodbfyTable.sql
```diff
@@ -897,7 +898,22 @@ BEGIN
...
+  -- but it's possible the source PK sequence is living in a name we would like to use
+  -- so we check to see if that's the case, and rename it out of the way
+  IF has_usable_primary_key AND has_usable_pk_sequence THEN
+    -- See if the existing sequence is squatting on our preferred name
+    destseq := Format('%s_%s_seq', relname, const.pkey);
+    SELECT pg_catalog.pg_get_serial_sequence(Format('%I.%I', relschema, relname), const.pkey)
+      INTO relseq;
+    -- If it's the name we want, then rename it
+    IF relseq IS NOT NULL AND relseq = Format('%I.%I', destschema, destseq) THEN
+      PERFORM _CDB_SQL(Format('ALTER SEQUENCE %s RENAME TO %s_tmp', relseq, destseq), '_CDB_Rewrite_Table');
```
jgoizueta: I must be missing something: it seems to me it's renaming the existing seq _relseq_ to the desired _destseq_ (<relname>_cartodb_id_seq) when it already has that very same name (?)

pramsey: It's moving `%s_cartodb_id_seq` to `_cartodb_id_seq_tmp` so that in the following steps, when the new table is created, it will be possible for the serial in `cartodb_id` to use `%s_cartodb_id_seq` as it's sequence name. Probably it would be best to recognize this case (sequence already there, and usable, even if the table is going to be re-written) as another case, but that would require larger changes below, to address what is a fairly minor corner case right now (the `ogr2ogr` upload issue).

jgoizueta: For some reason I was missing the `_tmp` 😅  thanks for the clarification!

oriolbx:
A client is experiencing this same issue when using ogr2ogr to have a ETL workflow to append data to the same table when importing with ogr2ogr. Having a message error like 

```
CARTO: RunSQL Response:{"error":["relation \"<relname>_cartodb_id_seq\" does not exist"]}

ERROR 1: Error returned by server : relation "<relname>_cartodb_id_seq" does not exist
```

rafatower:
Thank you guys for the fix and review.

Aside from the merge, there's also release and deploy for users to enjoy it. @oriolbx can you please reference the issue here and add as candidate to the RT?

oriolbx:
We gave the affected user an alternative to install and build the version where there is the fix. The fix went into gdal 2.2, so they can work off a stable branch https://trac.osgeo.org/gdal/changeset/41287
and to download it. They need to check out `svn checkout https://svn.osgeo.org/gdal/branches/2.2/ gdal-2.2-svn` and build the library

Issue #314, opened 2017-10-27
Slow performance in CDB_RectangleGrid


Algunenano:
Depending on the zoom level, `CDB_RectangleGrid` can be extremely slow. For example:

```sql
SELECT CDB_RectangleGrid ('0103000020110F000001000000050000003C36F772C5E369C19A6FECEB57AB43413C36F772C5E369C1292B3276739D5641D02678B00B995BC1292B3276739D5641D02678B00B995BC19A6FECEB57AB43413C36F772C5E369C19A6FECEB57AB4341'::geometry, CDB_XYZ_Resolution(14) * 12, CDB_XYZ_Resolution(14) * 12) as cell;
ERROR:  canceling statement due to statement timeout
```
```sql
SHOW statement_timeout;
 statement_timeout 
-------------------
 5min
(1 row)
```

The query was generated as part of the map preview and the full query looks something like this:
```sql
SELECT
    min(agg_value) min_val,
    max(agg_value) max_val,
    avg(agg_value) avg_val, 
    CDB_QuantileBins(array_agg(distinct(agg_value::numeric)), 5) as quantiles
FROM 
(
    WITH hgrid AS (
        SELECT CDB_RectangleGrid ( ST_Expand(ST_SetSRID(ST_GeomFromText('POLYGON((-13573560.9371348 2578210.49869538,-13573560.9371348 5928283.19126962,-7234721.4128759 5928283.19126962,-7234721.4128759 2578210.49869538,-13573560.9371348 2578210.49869538))'), 3857), CDB_XYZ_Resolution(14) * 12), CDB_XYZ_Resolution(14) * 12, CDB_XYZ_Resolution(14) * 12) as cell
    )
    SELECT
        hgrid.cell as the_geom_webmercator, count(1) as agg_value, count(1) / power( 12 * CDB_XYZ_Resolution(14), 2 ) as agg_value_density, row_number() over () as cartodb_id FROM hgrid, (SELECT * FROM yellow_1m) i where ST_Intersects(i.the_geom_webmercator, hgrid.cell) 
    GROUP BY hgrid.cell) _table_sql 

WHERE agg_value IS NOT NULL AND
    agg_value != 'infinity'::float AND
    agg_value != '-infinity'::float AND
    agg_value != 'NaN'::float;
```

Algunenano:
[...]

PR #322, opened 2018-02-14
Add limit to the grid-generating functions

jgoizueta:
This adds an optinal maxcells parameter to the rectangle and hexagonal grids,
with a default value of 262144 (2^18).

An error occurs when the grids would need more cell

### DIFF
--- a/scripts-available/CDB_Hexagon.sql
+++ b/scripts-available/CDB_Hexagon.sql
@@ -10,6 +10,10 @@ AS $$
     WHERE path[1] % 2 != 0
 $$ LANGUAGE 'sql' IMMUTABLE STRICT PARALLEL SAFE;
 
+
+-- In older versions of the extension, CDB_HexagonGrid had a different signature
+DROP FUNCTION IF EXISTS cartodb.CDB_HexagonGrid(GEOMETRY, FLOAT8, GEOMETRY);
+
 --
 -- Fill given extent with an hexagonal coverage
 --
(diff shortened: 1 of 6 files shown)


jgoizueta:
I've just realised I've spoiled parallel safety, right @Algunenano ?

jgoizueta:
This is used by Builder for aggregated styles; the kind of queries generated are (wrapping the original `$SOME_QUERY`:

```sql
WITH hgrid AS (
  SELECT CDB_RectangleGrid (
           ST_Expand(!bbox!, CDB_XYZ_Resolution($ZOOM) * $SIZE_PX),
           CDB_XYZ_Resolution($ZOOM) * $SIZE_PX,
           CDB_XYZ_Resolution($ZOOM) * $SIZE_PX
         ) as cell
)
SELECT
  hgrid.cell as the_geom_webmercator,
  count(1) as agg_value,
  count(1) /power( CDB_XYZ_Resolution($ZOOM) * $SIZE_PX, 2 ) as agg_value_density,
  row_number() over () as cartodb_id
FROM hgrid, ($SOME_QUERY) i
where ST_Intersects(i.the_geom_webmercator, hgrid.cell)
GROUP BY hgrid.cell
```

And we want to deal with the case that, because of the total extent of the dataset and the cell size (i.e. zoom level and size in pixels) a huge grid may be generated (producing timeouts or even out of memory errors).

@pramsey @Algunenano how important do you think that preserving parallelizability of the grid functions is here, with these use cases in mind?

jgoizueta:
Hey, has anybody removed this repo from Travis?

Algunenano:
I don't expect anyone needing multiple to make multiple parallel calls to CDB_RectangleGrid so I wouldn't consider it important; but if I have to check if raising an exception breaks it (catching them does, but I'm not sure about RAISE).

As I mentioned in https://github.com/CartoDB/cartodb-postgresql/issues/314 I don't like the approach taken here to generate the grids (generate all the grids and then join/intersect them with the dataset), but I don't know if there is a better way (I do hope so).

Also I'm wondering if the `maxcells` parameter added here should work as defined (raising an exception) or could otherwise work as a LIMIT of sorts and just only return the first maxcells generated; which also leads me to wonder whether LIMIT could work with pl/pgsql procedures returning sets. I'll give it more thought.

Travis seems really slow right now. Here is [the job](https://travis-ci.org/CartoDB/cartodb-postgresql/builds/341525079)

jgoizueta:
Thanks for the insights @Algunenano 
One possible way to to make this more efficient would be to use the aggregation API (with new placement modes to generate squares, and uhm... perhaps hexagons), but then the problem is that the current Builder gridding, unlike the original one in Editor, produces grids in which the cell size does not vary with the zoom level (so it is earth-bound rather than screen-bound as the size in pixels would seem to indicate). Our aggregation cell size is zoom-relative. I'm not sure how much sense it makes to add another aggregation mode in which the aggregation cell size is fixed in earth units.

jgoizueta:
I wonder why the tests fails (because of some extra CONTEXT message at the RAISEs).
The are ok in local, and I'm using `PGOPTIONS='-c client_min_messages=NOTICE'` as in travis.
Does that depend on PG version?

Algunenano:
Try adding something like `\set VERBOSITY TERSE` in the sql test file.

PR #332, opened 2018-07-03
Cdb table exists

rafatower:
This adds a new helper function `_CDB_Table_Exists(table_name_with_optional_schema TEXT)`. The intent is to use it as a guard for other calls. E.g:
```patch
diff --git a/app/services/carto/overviews_service.rb b/app/services/carto/overviews_service.rb
index 86211ff..eb7f067 100644
--- a/app/services/carto/overviews_service.rb
+++ b/app/services/carto/overviews_service.rb
@@ -26,11 +26,9 @@ module Carto
     end
 
     def delete_overviews(table_name)
-      # TODO: this is not very elegant, we could detect the existence of the
-      # table some otherway or have a CDB_DropOverviews(text)  function...
-      @database.run(%{SELECT cartodb.CDB_DropOverviews('#{table_name}'::REGCLASS)})
-    rescue Carto::Db::SqlInterface::Error => e
-      raise unless e.to_s.match /relation .+ does not exist/
+      if @database.fetch(%{SELECT cartodb._CDB_Table_Exists('#{table_name}')}).first[:_cdb_table_exists]
+        @database.run(%{SELECT cartodb.CDB_DropOverviews('#{table_name}'::REGCLASS)})
+      end
     end
 
     def overview_tables(table_name)
```

This way we avoid exception-driven code, and potentially breaking transactions.

Note that the error handling cannot be added to `cartodb.CDB_DropOverviews()`, because it fails as soon as some casting to `REGCLASS` is attempted on a non-existing table (i.e. it does not reach the body of the function, regardless of having the cast implicit or explicit).


scripts-available/CDB_Helper.sql
```diff
@@ -158,3 +158,21 @@ BEGIN
...
 $$ LANGUAGE 'plpgsql' IMMUTABLE PARALLEL SAFE;
+
+
+-- Checks if a given text representing a qualified or unqualified table name (relation)
+-- actually exists in the database. It is meant to be used as a guard for other function/queries.
+CREATE OR REPLACE FUNCTION cartodb._CDB_Table_Exists(table_name_with_optional_schema TEXT)
+RETURNS bool
+AS $$
+BEGIN
+    IF EXISTS(SELECT * FROM pg_class WHERE table_name_with_optional_schema::regclass::oid = oid AND relkind = 'r') THEN
```
Algunenano: Can this be done without the `if`? That is assigning it to a variable and returning it plus catching the exception. I'm asking without knowing if its possible in plpgsql but it feels more natural.

scripts-available/CDB_Helper.sql
```diff
@@ -158,3 +158,21 @@ BEGIN
...
+       RETURN true;
+    ELSE
+       RETURN false;
+    END IF;
+EXCEPTION
+    WHEN invalid_schema_name OR undefined_table THEN
+        RETURN false;
+END;
+$$ LANGUAGE PLPGSQL VOLATILE PARALLEL SAFE;
```
Algunenano: If the function catches exceptions it should usually be declared a PARALLEL UNSAFE [[ref](https://www.postgresql.org/docs/10/static/parallel-safety.html)]:
`Functions and aggregates must be marked PARALLEL UNSAFE if they write to the database, access sequences, change the transaction state even temporarily (e.g. a PL/pgSQL function which establishes an EXCEPTION block to catch errors),`

scripts-available/CDB_Helper.sql
rafatower: Actually I think that's the recommended construction for it, so that no extra variable (plus check) is needed.

Issue #341, opened 2018-10-10
Add DROP function to the artifact in order to avoid migration problems


ethervoid:
When we create a new version we don't add the DROP to functions that have been removed creating a problem when we try to migrate clients with those old functions.

Algunenano:
Unless we've forgotten about it somewhere we are already doing this ([example](https://github.com/CartoDB/cartodb-postgresql/pull/340/files#diff-abaa85f717a1afdd6b0680a776aed39aR149).

Do you have a list of troublesome functions so we add the drop?

ethervoid:
Yep, old ones. We have migration problems with the DDL old functions:

```
cdb_handle_create_table()
cdb_handle_drop_table()
cdb_handle_alter_column()
cdb_handle_drop_column()
cdb_handle_add_column()
cdb_disable_ddl_hooks()
cdb_enable_ddl_hooks()
```

I'm pretty sure they're not in the current versions right now but this ticket was more like a suggestion if we're not doing it :)

javitonino:
cc @CartoDB/rt-managers (we have replaced #343 for this one)

PR #344, opened 2019-01-02
Recover hooks

Algunenano:
References https://github.com/CartoDB/cartodb-postgresql/issues/341

Travis build was broken (once again), so I've updated it to pass with PG11 too. With this update `CDB_TransformToWebmercator` broke once again so now it handles yet another corner case.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,38 @@
+dist: xenial
 language: c
-dist: trusty
 sudo: required
 
 env:
-  - PGUSER=postgres PGDATABASE=postgres
+  global:
+    - PGUSER=postgres
+    - PGDATABASE=postgres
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,26 +1,38 @@
...
-  - PGUSER=postgres PGDATABASE=postgres
+  global:
+    - PGUSER=postgres
+    - PGDATABASE=postgres
+    - PGOPTIONS='-c client_min_messages=NOTICE'
+    - PGPORT=5432
+    - POSTGIS_VERSION="2.5"
+
+  matrix:
+    - POSTGRESQL_VERSION="9.5"
```
javitonino: Maybe we can drop 9.5 now?

Algunenano: Maybe in the future. Since it's working right now I'll leave it for now

PR #346, opened 2019-02-20
Add CDB_Username function

gonzaloriestra:
Added `CDB_Username()` to get the cartodb username from the current PostgreSQL user.

I had to create two nested functions:
* `CDB_Username()`: retrieves the current_user (usually not superuser) and calls the next one with it
* `_CDB_Username(pg_user)`: access to `CDB_Conf` with superuser rights and retrieve the username from there.

The problem of this approach is that any user could use `_CDB_Username(pg_user)` to get the username of other users if they know the PostgreSQL user. But I couldn't find a way to make it private and I guess it's acceptable.


Algunenano:
Have you considered enabling [Row Security Policies](https://www.postgresql.org/docs/11/ddl-rowsecurity.html) in `cdb_conf` and giving permission to read only certain rows?

gonzaloriestra:
@Algunenano Thanks, I didn't know it. I think it might work, but we would have to do some changes in CartoDB to update the permissions when we create/destroy users and also to update them for the existing users. So I'm not sure it's worth trying it... What do you think @javitonino?

Algunenano:
In a simpler way. Would using `session_user` be enough?
```sql
CREATE OR REPLACE FUNCTION CDB_Username()
RETURNS text
AS $$
  SELECT CDB_Conf_GetConf(CONCAT('api_keys_', session_user))->>'username';
$$ LANGUAGE SQL STRICT IMMUTABLE PARALLEL SAFE SECURITY DEFINER;
```

javitonino:
Uhm, that's an intriguing idea @Algunenano.

A proof of concept:
```sql
ALTER TABLE cdb_conf ENABLE ROW LEVEL SECURITY;
GRANT SELECT ON cdb_conf TO PUBLIC;
CREATE POLICY own_api_key ON cdb_conf FOR SELECT USING (key = 'api_keys_' || current_user);
```

I feel super-afraid giving SELECT to PUBLIC here, so it works perfectly fine.

The `session_user` thing might be simpler though and it should work. Since we don't really support switching roles in the middle of a session (and in any case, not between roles from different CARTO users), it should be reliable.

gonzaloriestra:
The `session_user` trick works, thanks! :slightly_smiling_face:

Issue #347, opened 2019-02-26
[Ghost tables] Event trigger for DDL instructions


ethervoid:
A new function that will be used to encapsulate all the even trigger logic. We have to decide if we leave the event trigger outside of this

- [ ] move SQL code from `create_ghost_tables_event_trigger` to `cartodb-postgresql` extension
- [ ] use TIS to enque the JOB

ethervoid:
Some thoughts for this new trigger and its implications:

We have to be careful with the DDL instructions we process. Why? because right now we trigger the Ghost Tables when the user enters in the dashboard but in the future, we're going to launch the events when a DDL instruction is executed. Why is that a problem? Because we're going to enqueue more jobs, I'm thinking in alter table with drop column for example, than before in shared queues and that could delay other unrelated tasks so maybe we want to filter some DDL instructions.

Another thing to take into account is the Importer issue. It's highly probable that things are going to mess up when the importer starts to enqueue ghost tables at the same time all the importer operations are being done. So maybe have to lock GT for that user using Bolt until the importer finishes and execute GT after that.

PR #349, opened 2019-03-06
Ghost tables functions

gonzaloriestra:
Closes https://github.com/CartoDB/cartodb-postgresql/issues/347


scripts-available/CDB_GhostTables.sql
```diff
@@ -0,0 +1,104 @@
...
+$$ LANGUAGE plpgsql VOLATILE PARALLEL UNSAFE SECURITY DEFINER;
+
+-- Creates the trigger on DDL events to link ghost tables
+CREATE OR REPLACE FUNCTION CDB_EnableGhostTablesTrigger()
+RETURNS void
+AS $$
+  BEGIN
+    CREATE EVENT TRIGGER link_ghost_tables
+    ON ddl_command_end
+    WHEN TAG IN ('CREATE TABLE', 'SELECT INTO', 'DROP TABLE', 'ALTER TABLE', 'CREATE TRIGGER', 'DROP TRIGGER')
```
ethervoid: Add CREATE/DROP/ALTER VIEW

scripts-available/CDB_GhostTables.sql
```diff
@@ -0,0 +1,104 @@
...
+$$ LANGUAGE 'plpythonu' VOLATILE PARALLEL UNSAFE;
+
+-- Enqueues a job to run Ghost tables linking process for the current user
+CREATE OR REPLACE FUNCTION CDB_LinkGhostTables()
+RETURNS void
+AS $$
+  DECLARE
+    user_id TEXT;
+  BEGIN
+    EXECUTE 'SELECT (regexp_match(session_user, ''[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}''))[1];' INTO user_id;
```
ethervoid: I've been thinking about this and maybe, to avoid the problems derived from the user of regexp (for example this could not work with regular apis), we should send the username to the TIS and create another Job in cartodb that:
- Get the ID from username
- Call the ghost table task with that ID

scripts-available/CDB_GhostTables.sql
```diff
@@ -0,0 +1,104 @@
...
+    INSERT INTO cartodb.cdb_ddl_execution VALUES (txid_current()) ON CONFLICT (txid) DO NOTHING;
+  END;
+$$ LANGUAGE plpgsql VOLATILE PARALLEL UNSAFE SECURITY DEFINER;
+
+-- Creates the trigger on DDL events to link ghost tables
+CREATE OR REPLACE FUNCTION CDB_EnableGhostTablesTrigger()
+RETURNS void
+AS $$
+  BEGIN
+    CREATE EVENT TRIGGER link_ghost_tables
```
ethervoid: We should do here a DROP IF EXISTS and a CREATE  for the trigger so we can use this function to upgrade the trigger in case of need

scripts-available/CDB_GhostTables.sql
```diff
@@ -0,0 +1,116 @@
...
+      client.execute_command('DBSCH', db_name, username, ddl_tag)
+      break
+    except Exception as err:
+      error = "request_error - %s" % str(err)
+      client = GD['invalidation'] = None # force reconnect
+      if not tis_retry:
+        plpy.warning('Invalidation Service error: ' +  str(err))
+        break
+      tis_retry -= 1 # try reconnecting
+$$ LANGUAGE 'plpythonu' VOLATILE PARALLEL UNSAFE;
```
ethervoid: Maybe we could swap the `SECURITY DEFINER` from the `public` function to here because this function is the one using the admin rights to access cdb_conf

scripts-available/CDB_GhostTables.sql
```diff
@@ -0,0 +1,116 @@
...
+  tis_retry = tis_config_dict.get('retry', 5)
+      
+  client = GD.get('invalidation', None)
+
+  while True:
+
+    if not client:
+        try:
+          import redis
+          client = GD['invalidation'] = redis.Redis(host=tis_host, port=tis_port, socket_timeout=tis_timeout)
```
ethervoid: This is a bit ugly. I'd use two lines for this.

scripts-available/CDB_GhostTables.sql
gonzaloriestra: We also need it in the public one to access `cdb_ddl_execution`.

ethervoid: Ok, i see

scripts-available/CDB_GhostTables.sql
```diff
@@ -0,0 +1,117 @@
...
+RETURNS trigger
+AS $$
+  BEGIN
+    PERFORM CDB_LinkGhostTables();
+    RETURN NULL;
+  END;
+$$ LANGUAGE plpgsql VOLATILE PARALLEL UNSAFE SECURITY DEFINER;
+
+-- Trigger to call CDB_LinkGhostTables() when adding a row in cartodb.cdb_ddl_execution
+DROP TRIGGER IF EXISTS check_ddl_update ON cartodb.cdb_ddl_execution;
```
ethervoid: This trigger creation should be wrapped as well, if not is going to be created/installed always

scripts-available/CDB_GhostTables.sql
```diff
@@ -0,0 +1,117 @@
...
+    INSERT INTO cartodb.cdb_ddl_execution VALUES (txid_current(), tg_tag) ON CONFLICT (txid) DO NOTHING;
+  END;
+$$ LANGUAGE plpgsql VOLATILE PARALLEL UNSAFE SECURITY DEFINER;
+
+-- Creates the trigger on DDL events to link ghost tables
+CREATE OR REPLACE FUNCTION CDB_EnableGhostTablesTrigger()
+RETURNS void
+AS $$
+  BEGIN
+    DROP EVENT TRIGGER IF EXISTS link_ghost_tables;
```
ethervoid: Did you check if hte `IF EXISTS` works here?

scripts-available/CDB_GhostTables.sql
```diff
@@ -0,0 +1,117 @@
...
+  if not username:
+    return
+
+  if 'json' not in GD:
+    import json
+    GD['json'] = json
+  else:
+    json = GD['json']    
+
+  error = ''
```
ethervoid: Why are you declaring this variable here empty?

scripts-available/CDB_GhostTables.sql
```diff
@@ -0,0 +1,117 @@
...
+  tis_config = plpy.execute("select cartodb.CDB_Conf_GetConf('invalidation_service');")[0]['cdb_conf_getconf']
+  tis_config_dict = json.loads(tis_config) if tis_config else {}
+  tis_host = tis_config_dict.get('host', '127.0.0.1')
+  tis_port = tis_config_dict.get('port', 3142)
+  tis_timeout = tis_config_dict.get('timeout', 5)
+  tis_retry = tis_config_dict.get('retry', 5)
+      
+  client = GD.get('invalidation', None)
+
+  while True:
```
ethervoid: Ph34r!

scripts-available/CDB_GhostTables.sql
ethervoid: Maybe is a good idea do it inside the `CDB_EnableGhostTablesTrigger` function and drop it in the disable one

scripts-available/CDB_GhostTables.sql
gonzaloriestra: Yes, it works :+1:

scripts-available/CDB_GhostTables.sql
gonzaloriestra: Copied from another place. Removing it as it's not needed...

scripts-available/CDB_GhostTables.sql
gonzaloriestra: I don't like it either, but it comes from a similar function and it works. Do you want me to use `this_retry` here instead?

scripts-available/CDB_GhostTables.sql
gonzaloriestra: I left it outside because it wasn't going to be used without the event trigger enabled, but yeah, it makes sense what you say. Let's do it!

scripts-available/CDB_GhostTables.sql
ethervoid: I'm not sure, that's why I put ph34r. We can let that as it is

scripts-available/CDB_GhostTables.sql
gonzaloriestra: I'm thinking that we could even create/drop the `cdb_ddl_execution` in the functions, but it would require some additional work on `CDB_LinkGhostTables()` to avoid crashing without the trigger enabled. What do you think?

ethervoid: I was going to write that here hahaha. I"ll create the table inside the function too and drop it in case we disable the trigger. But at least I'll create the table inside the function so if we need to upgrade it we can do operations inside those wrappers. What do you think?

gonzaloriestra: I've finally moved it to the functions too as we have talked.

PR #350, opened 2019-03-12
Add explicit schema for Ghost Tables functions

gonzaloriestra:
It's safer to always use the cartodb schema for the functions. For example, some tests are failing in cartodb when installing this new trigger because it doesn't find the functions (because search_path is modified).

### DIFF
--- a/scripts-available/CDB_GhostTables.sql
+++ b/scripts-available/CDB_GhostTables.sql
@@ -53,7 +53,7 @@ AS $$
     username TEXT;
     db_name TEXT;
   BEGIN
-    EXECUTE 'SELECT CDB_Username();' INTO username;
+    EXECUTE 'SELECT cartodb.CDB_Username();' INTO username;
     EXECUTE 'SELECT current_database();' INTO db_name;
 
     PERFORM cartodb._CDB_LinkGhostTables(username, db_name, event_name);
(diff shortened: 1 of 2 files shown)


scripts-available/CDB_GhostTables.sql
```diff
@@ -56,27 +56,26 @@ AS $$
...
+CREATE OR REPLACE FUNCTION cartodb._CDB_LinkGhostTablesTrigger()
 RETURNS trigger
 AS $$
   DECLARE
     ddl_tag TEXT;
   BEGIN
-    EXECUTE 'SELECT tag FROM cartodb.cdb_ddl_execution WHERE txid = txid_current();' INTO ddl_tag;
-    DELETE FROM cartodb.cdb_ddl_execution WHERE txid = txid_current();
-    PERFORM CDB_LinkGhostTables(ddl_tag);
+    EXECUTE 'DELETE FROM cartodb.cdb_ddl_execution WHERE txid = txid_current() RETURNING tag;' INTO ddl_tag;
```
oleurud: :top:

scripts-available/CDB_GhostTables.sql
```diff
@@ -56,27 +56,26 @@ AS $$
     EXECUTE 'SELECT CDB_Username();' INTO username;
```
oleurud: do we need `cartodb` here?

oleurud: Yes, we need it

PR #351, opened 2019-03-14
Remove default TIS config in Ghost Table function

gonzaloriestra:
Remove default values for TIS connection to avoid failing attempts when it hasn't been configured.

### DIFF
--- a/scripts-available/CDB_GhostTables.sql
+++ b/scripts-available/CDB_GhostTables.sql
@@ -13,6 +13,7 @@ AS $$
 
   tis_config = plpy.execute("select cartodb.CDB_Conf_GetConf('invalidation_service');")[0]['cdb_conf_getconf']
   if not tis_config:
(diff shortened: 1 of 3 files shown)


scripts-available/CDB_GhostTables.sql
```diff
@@ -12,9 +12,12 @@ AS $$
     json = GD['json']    
 
   tis_config = plpy.execute("select cartodb.CDB_Conf_GetConf('invalidation_service');")[0]['cdb_conf_getconf']
-  tis_config_dict = json.loads(tis_config) if tis_config else {}
-  tis_host = tis_config_dict.get('host', '127.0.0.1')
-  tis_port = tis_config_dict.get('port', 3142)
+  if not tis_config:
+    return
```
oleurud: It could be useful to have a warning here: "Add Invalidation Service configuration to link Ghost Tables"

PR #355, opened 2019-05-27
[WIP] Table syncer

rafatower:
It assumes there's a cartodb_id column in both source and target. It
does not perform unnecessary actions. It respects augmented columns in
target table, if they exist. It is meant to be efficient.

### DIFF
--- a/scripts-available/CDB_SyncTable.sql
+++ b/scripts-available/CDB_SyncTable.sql
@@ -52,6 +52,22 @@ BEGIN
 END;
 $$ LANGUAGE plpgsql IMMUTABLE PARALLEL SAFE;
 
+
+/*
+    Given a prefix, generate a safe unique NAME for a temp table.
+
+    Example of usage:
+
+       SELECT __CDB_GenerateUniqueName('src_sync'); --> src_sync_718794_120106
+
+*/
+CREATE OR REPLACE FUNCTION __CDB_GenerateUniqueName(prefix TEXT)
+RETURNS NAME
+AS $$
+  SELECT format('%s_%s_%s', prefix, txid_current(), (random()*1000000)::int)::NAME;
+$$ LANGUAGE sql VOLATILE PARALLEL UNSAFE;
+
+
 /*
    A Table Syncer
 
@@ -94,8 +110,8 @@ BEGIN
   SELECT ARRAY(SELECT quote_ident(c) FROM _CDB_GetColumns(src_table) as c WHERE c <> 'cartodb_id') INTO colnames;
   quoted_colnames := array_to_string(colnames, ',');
 
-  src_hash_table_name := format('src_sync_%s', txid_current());
-  dst_hash_table_name := format('dst_sync_%s', txid_current());
+  src_hash_table_name := __CDB_GenerateUniqueName('src_sync');
+  dst_hash_table_name := __CDB_GenerateUniqueName('dst_sync');
 
   EXECUTE format('CREATE TEMP TABLE %I(cartodb_id BIGINT, hash TEXT) ON COMMIT DROP', src_hash_table_name);
   EXECUTE format('CREATE TEMP TABLE %I(cartodb_id BIGINT, hash TEXT) ON COMMIT DROP', dst_hash_table_name);
--- a/scripts-available/CDB_SyncTable.sql
+++ b/scripts-available/CDB_SyncTable.sql
@@ -3,9 +3,9 @@
 
    Sample usage:
 
-     SELECT _CDB_GetColumns('public.films');
+     SELECT cartodb._CDB_GetColumns('public.films');
 */
-CREATE OR REPLACE FUNCTION _CDB_GetColumns(src_table REGCLASS)
+CREATE OR REPLACE FUNCTION cartodb._CDB_GetColumns(src_table REGCLASS)
 RETURNS SETOF NAME
 AS $$
   SELECT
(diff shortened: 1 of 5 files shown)


rafatower:
@jgoizueta although this is pretty complete now, I'd like to try and integrate it w/ Geocodings before going any further. There are things that may impact the function and its signature that I cannot clearly foresee yet. E.g: should it accept an array of columns to exclude? `the_geom` and `the_geom_webmercator`?

I also want to give us some time to think about this in different contexts: CartoFrames, Geocodings and Import API.

cc/ @javitonino

rafatower:
The test failure in PG11 :disappointed: 

```
$ cat regression.diffs
*** /home/travis/build/CartoDB/cartodb-postgresql/expected/test/CDB_AnalysisCheckTest.out	2019-05-27 16:09:45.063543994 +0000
--- /home/travis/build/CartoDB/cartodb-postgresql/results/test/CDB_AnalysisCheckTest.out	2019-05-27 16:12:39.770847666 +0000
***************
*** 5,12 ****
  CREATE TABLE
  CREATE TABLE
  CREATE TABLE
- (analysis_2f13a3dbd7_41bd92976fc6dd97072afe4ee450054f4c0715d5,public,analysis_2f13a3dbd7_41bd92976fc6dd97072afe4ee450054f4c0715d5)
  (analysis_2f13a3dbd7_f00cee44e9e6152b450bde3a92eb9ae0d099da94,public,analysis_2f13a3dbd7_f00cee44e9e6152b450bde3a92eb9ae0d099da94)
  0
  
  1
--- 5,12 ----
  CREATE TABLE
  CREATE TABLE
  CREATE TABLE
  (analysis_2f13a3dbd7_f00cee44e9e6152b450bde3a92eb9ae0d099da94,public,analysis_2f13a3dbd7_f00cee44e9e6152b450bde3a92eb9ae0d099da94)
+ (analysis_2f13a3dbd7_41bd92976fc6dd97072afe4ee450054f4c0715d5,public,analysis_2f13a3dbd7_41bd92976fc6dd97072afe4ee450054f4c0715d5)
  0
  
  1
```

jgoizueta:
For the geocoding we'll need to it to ignore `the_geom` and `the_geom_webmercator` from the source, so I think that having an array of columns to ignore in the source would be a good solution.

Besides integrating this into the geocoding I'll  try to think about other use cases to look for other requirements.

This looks great!

scripts-available/CDB_SyncTable.sql
```diff
@@ -0,0 +1,150 @@
...
+  END IF;
+
+  -- Get the list of columns from the source table, excluding cartodb_id
+  SELECT ARRAY(SELECT quote_ident(c) FROM _CDB_GetColumns(src_table) as c WHERE c <> 'cartodb_id') INTO colnames;
+  quoted_colnames := array_to_string(colnames, ',');
+
+  src_hash_table_name := format('src_sync_%s', txid_current());
+  dst_hash_table_name := format('dst_sync_%s', txid_current());
+
+  EXECUTE format('CREATE TEMP TABLE %I(cartodb_id BIGINT, hash TEXT) ON COMMIT DROP', src_hash_table_name);
```
jgoizueta: It could be very messy, but have you considered persisting the source hash table  (to be used as the next destination hash table)?

jgoizueta: Oh, I realize it's a very bad idea, regarding concurrent changes etc.

rafatower: Really not worth it, as per the few "load tests" I performed in my machine. The whole process can take less than 30s for a 1.5M rows table with changes in 10% of the rows, so... Also, the whole process is transactional because of the way PG functions work:

Up to PG10:

https://www.postgresql.org/docs/10/plpgsql-structure.html

> Functions and trigger procedures are always executed within a transaction established by an outer query — they cannot start or commit that transaction, since there would be no context for them to execute in

In PG11:

https://www.postgresql.org/about/news/1894/

> PostgreSQL 11 adds SQL procedures that can perform full transaction management within the body of a function, enabling developers to create more advanced server-side applications, such as ones involving incremental bulk data loading.

https://www.postgresql.org/docs/11/plpgsql-transactions.html

jgoizueta:
I would like to test some variations with some big tables and compare the performance of each of the three phases (DELETE, INSERT, UPDATE).

Operating without the tmp tables: (⚠️ buggy sketch of code following)

```sql
 -- Deal with deleted rows: ids in dest but not in source
  EXECUTE format(
    'DELETE FROM %s _dst WHERE NOT EXISTS (SELECT * FROM %I _src WHERE _src.cartodb_id=_dst.cartodb_id)',
    fq_dest_table,
    src_table);
  GET DIAGNOSTICS num_rows = ROW_COUNT;
  RAISE NOTICE 'DELETED % row(s)', num_rows;

   -- Deal with inserted rows: ids in source but not in dest
  EXECUTE format('
      INSERT INTO %s _dst (cartodb_id,%s)
      SELECT cartodb_id,%s FROM %I _src WHERE NOT EXISTS (SELECT * FROM _dst WHERE _src.cartodb_id=_dst.cartodb_id)
  ', fq_dest_table, quoted_colnames, quoted_colnames, src_table);
  GET DIAGNOSTICS num_rows = ROW_COUNT;
  RAISE NOTICE 'INSERTED % row(s)', num_rows;

   -- Deal with modified rows
  update_set_clause := __CDB_GetUpdateSetClause(colnames, 'changed');
  EXECUTE format('
    UPDATE %1$s dst SET %2$s
    FROM (
      SELECT * FROM %3$s src
      -- WHERE  md5(ROW(src.*)::text) <> md5(ROW(dst.*)::text)
    ) changed
    WHERE dst.cartodb_id = changed.cartodb_id;
  ', fq_dest_table, update_set_clause, src_table);
  GET DIAGNOSTICS num_rows = ROW_COUNT;
  RAISE NOTICE 'MODIFIED % row(s)', num_rows;
```

And, for the modified rows, compare using the hash comparison vs just updating all the rows (commented WHERE).

PR #356, opened 2019-05-30
COALESCE source columns

rafatower:
What it does: it changes the generated UPDATE SET caluse so that
instead of

  the_geom = changed.the_geom

it will write

  the_geom = COALESCE(changed.the_geom, the_geom)

So, if the source column has a NULL value, it will keep whatever value
the target column has for that row.

This may be a general mechanism to work around syncing in a general
case. We may decide to apply it to specific columns instead but that
complicates the implementation (instead of skip_columns it may have
a harder-to-understand param "coalesce_columns", *if* they exist).

I'm creating this to discuss it. Note it greatly affects the simple
API.

### DIFF
--- a/scripts-available/CDB_SyncTable.sql
+++ b/scripts-available/CDB_SyncTable.sql
@@ -29,15 +29,15 @@ $$ LANGUAGE sql STABLE PARALLEL UNSAFE;
     Given an array of quoted column names, it generates an UPDATE SET
     clause with the following form:
 
-        the_geom = changed.the_geom,
-        id = changed.id,
-        elevation = changed.elevation
+        the_geom = COALESCE(changed.the_geom, dst.the_geom),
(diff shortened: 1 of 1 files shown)


rafatower:
Note: in a quick test, it did not seem to affect performance (not noticeably).

cc/ @javitonino

jgoizueta:
I think we'd need some control over this (some parametrization), since it applies only to some use cases. For example, this could be OK to sync a table before geocoding, but not to sync the geocoded table back to a public table.
For general sync usage I have my doubts: it could probably work for some cases, but in others NULL can be meaningful and should be synced.

rafatower:
Fair enough. Okay, let's discard this, as it may have undesired consequences and I can picture my future self struggling to understand convoluted sync scenarios.

PR #357, opened 2019-05-31
Qualify all function calls

Algunenano:
* Fully qualify function calls
* Several improvements to bash tests.
* Avoid dropping publicuser in tests.
* Raise minimum requirement to PostgreSQL 9.6.

Pending:
- [x] Test that it doesn't break builder: Works locally
- [x] Test pg_upgrade (broken right now when an index is created over `cdb_latlng`: Confirmed that it's fixed.
- [ ] Adapt Windshaft / Windshaft-cartodb / SQL-API to the new ways: either patch the SQL files (as it was done with PARALLEL) or try to stop downloading them directly.
- [ ] Upgrade.

Conflicts with https://github.com/CartoDB/cartodb-postgresql/pull/355 (some changes there will be necessary once both are merged).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ env:
     - POSTGIS_VERSION="2.5"
 
   matrix:
-    - POSTGRESQL_VERSION="9.5"
     - POSTGRESQL_VERSION="10"
     - POSTGRESQL_VERSION="11"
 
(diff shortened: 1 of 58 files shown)


.travis.yml
```diff
@@ -11,7 +11,6 @@ env:
     - POSTGIS_VERSION="2.5"
 
   matrix:
-    - POSTGRESQL_VERSION="9.5"
```
rafatower: replace by 9.6 as per changelog?

rafatower:
> Avoid dropping publicuser in tests

:heart_eyes: 

With this kind of massive changes I see little value in visual inspetion. I'd rather review the process (e.g: if you applied a sed script or similar, the looks of the regular expressions and the logic behind) and rely on the automated tests.

Anyway it will require at least a manual upgrade to check everything's ok.

> Conflicts with #355

No worries with that one. No rush and will be adapted before merging. It's still WIP as we don't have the function signature fully closed yet. Thanks for the heads up

Algunenano:
> I'd rather review the process

I tried automating it, ended up reviewing and doing most of it by hand.

I've tried this locally and it works fine with builder and it fixes the pg_upgrade issue (index over cdb_latlng) so I'm going to add some information to CONTRIBUTING.md explaining how to use the schemas and the merge this and check the other 2 pending tasks:
- Pass builder tests.
- Fix Windshaft/SQL-API since they download the files directly and they require parsing those 2 variables (as they used to do with PARALLEL tags).

PR #363, opened 2019-07-01
Fix cartodb schema references

jgoizueta:
Use @extschema@ for the schema of the extension

### DIFF
--- a/scripts-available/CDB_SyncTable.sql
+++ b/scripts-available/CDB_SyncTable.sql
@@ -3,9 +3,9 @@
 
    Sample usage:
 
-     SELECT cartodb._CDB_GetColumns('public.films');
+     SELECT @extschema@._CDB_GetColumns('public.films');
 */
-CREATE OR REPLACE FUNCTION cartodb._CDB_GetColumns(src_table REGCLASS)
+CREATE OR REPLACE FUNCTION @extschema@._CDB_GetColumns(src_table REGCLASS)
 RETURNS SETOF NAME
 AS $$
   SELECT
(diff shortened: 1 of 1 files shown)


scripts-available/CDB_SyncTable.sql
```diff
@@ -35,9 +35,9 @@ $$ LANGUAGE sql STABLE PARALLEL UNSAFE;
 
     Example of usage:
 
-       SELECT cartodb.__CDB_GetUpdateSetClause('{the_geom, id, elevation}', 'changed');
+       SELECT @extschema@.__CDB_GetUpdateSetClause('{the_geom, id, elevation}', 'changed');
```
Algunenano: You probably can leave the comment as it was.

scripts-available/CDB_SyncTable.sql
```diff
@@ -81,7 +81,7 @@ $$ LANGUAGE sql VOLATILE PARALLEL UNSAFE;
      SELECT cartodb.CDB_SyncTable('test_sync_source', 'public', 'test_sync_dest', '{the_geom, the_geom_webmercator}');
 
 */
-CREATE OR REPLACE FUNCTION cartodb.CDB_SyncTable(src_table REGCLASS, dst_schema REGNAMESPACE, dst_table NAME, skip_cols NAME[] = '{}')
+CREATE OR REPLACE FUNCTION @extschema@.CDB_SyncTable(src_table REGCLASS, dst_schema REGNAMESPACE, dst_table NAME, skip_cols NAME[] = '{}')
```
Algunenano: Any reason for using `schema.CDB_SyncTable` instead of just `CDB_SyncTable` and call it normally?

jgoizueta: I think it'll be better to leave out the schema for a *public* function:

PR #365, opened 2019-07-02
Sync tables without using temporary hash tables

jgoizueta:
Avoid the usage of the tmp tables.
Performance seems similar. Should be thoroughly tested.


scripts-available/CDB_SyncTable.sql
```diff
@@ -88,7 +101,8 @@ DECLARE
   fq_dest_table TEXT;
 
   colnames TEXT[];
-  quoted_colnames TEXT;
+  dst_colnames TEXT;
+  src_colnames TEXT;
 
   src_hash_table_name NAME;
   dst_hash_table_name NAME;
```
rafatower: these two `src_hash_table_name` and `dst_hash_table_name` can be deleted.

scripts-available/CDB_SyncTable.sql
```diff
@@ -67,6 +67,19 @@ AS $$
...
+    return array of column names qualified with the table name
+
+    Example of usage:
+
+       SELECT cartodb.__CDB_QualifyColumns('t', ARRAY['a','b']); --> ARRAY['t.a','t.b']
+
+*/
+CREATE OR REPLACE FUNCTION cartodb.__CDB_QualifyColumns(tablename TEXT, colnames TEXT[]) RETURNS TEXT[] AS
+$$
+    SELECT array_agg(tablename || '.' || _colname) from unnest(colnames) _colname;
```
rafatower: I think this needs a couple `quote_ident()` to escape identifiers when needed.

rafatower: Not sure, though and beware of double-quoting:

```
select quote_ident('Hola');
 quote_ident 
-------------
 "Hola"
(1 row)
```

```
select quote_ident(quote_ident('Hola'));
 quote_ident 
-------------
 """Hola"""
(1 row)
```

Also mind that `NAME` type is aimed at holding identifiers (with it's infamous 63 bytes limit).

scripts-available/CDB_SyncTable.sql
jgoizueta: true

PR #367, opened 2019-07-03
Oauth create table trigger

ethervoid:
Related https://github.com/CartoDB/cartodb/issues/14594

Event trigger that reassigns the owner of the created table to the owner of the oauth app


scripts-available/CDB_OAuth.sql
```diff
@@ -0,0 +1,53 @@
...
+    FOR obj IN SELECT * FROM pg_event_trigger_ddl_commands()
+    LOOP
+      RAISE DEBUG '% ddl object: % % % %',
+                  tg_tag,
+                  obj.command_tag,
+                  obj.object_type,
+                  obj.schema_name,
+                  obj.object_identity;
+      SELECT rolname FROM pg_class JOIN pg_roles ON relowner = pg_roles.oid WHERE pg_class.oid = obj.objid INTO creator_role;
+      SELECT value->>'ownership_role_name' from cdb_conf where key = (SELECT 'api_keys_' || rolname FROM pg_class JOIN pg_roles ON relowner = pg_roles.oid WHERE pg_class.oid = obj.objid) INTO owner_role;
```
alrocar: Could we reuse `creator_role`?

```suggestion
      SELECT value->>'ownership_role_name' from cdb_conf where key = 'api_keys_' || creator_role INTO owner_role;
```

scripts-available/CDB_OAuth.sql
```diff
@@ -0,0 +1,53 @@
...
+-- Creates the trigger on DDL events in order to reassign the owner
+CREATE OR REPLACE FUNCTION @extschema@.CDB_EnableOAuthReassignTablesTrigger()
+RETURNS void
+AS $$
+  BEGIN
+    DROP EVENT TRIGGER IF EXISTS oauth_reassign_tables_trigger;
+
+    CREATE EVENT TRIGGER oauth_reassign_tables_trigger
+    ON ddl_command_end
+    WHEN TAG IN ('CREATE TABLE', 'CREATE TABLE AS', 'SELECT INTO', 'CREATE VIEW', 'CREATE FOREIGN TABLE', 'CREATE MATERIALIZED VIEW')
```
alrocar: Should we include here `CREATE OR REPLACE TABLE/VIEW, etc.`?

test/CDB_OAuth.sql
```diff
@@ -0,0 +1,48 @@
...
+
+CREATE TABLE test2(id INT);
+INSERT INTO test2 VALUES(1);
+SELECT * FROM test2;
+
+\set QUIET on
+SET SESSION AUTHORIZATION "ownership_role";
+\set QUIET off
+
+SELECT * FROM test2;
```
alrocar: After this I'd check that `creator_role` can drop/alter the table.

scripts-available/CDB_OAuth.sql
ethervoid: :eye:

ethervoid: No need, the DDL instruction is `CREATE VIEW` the `OR REPLACE` is optional

scripts-available/CDB_OAuth.sql
```diff
@@ -0,0 +1,53 @@
...
+    LOOP
+      RAISE DEBUG '% ddl object: % % % %',
+                  tg_tag,
+                  obj.command_tag,
+                  obj.object_type,
+                  obj.schema_name,
+                  obj.object_identity;
+      SELECT rolname FROM pg_class JOIN pg_roles ON relowner = pg_roles.oid WHERE pg_class.oid = obj.objid INTO creator_role;
+      SELECT value->>'ownership_role_name' from cdb_conf where key = (SELECT 'api_keys_' || rolname FROM pg_class JOIN pg_roles ON relowner = pg_roles.oid WHERE pg_class.oid = obj.objid) INTO owner_role;
+      IF owner_role IS NULL THEN
```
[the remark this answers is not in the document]
ethervoid: yup, adding an empty string as well. goo catch

test/CDB_OAuth.sql
ethervoid: Done and added more tests to check also for null/empty `cdb_conf` config

alrocar: ❤️

test/CDB_OAuth.sql
```diff
@@ -0,0 +1,175 @@
...
+SET SESSION AUTHORIZATION "ownership_role";
+\set QUIET off
+
+SELECT * FROM test4;
+SELECT * FROM test4_tablesas;
+SELECT * FROM test4_view;
+SELECT * FROM test4_mview;
+SELECT * FROM test4_selectinto;
+
+-- Ownership role drops the tables
```
alrocar: In the end, an app is going to run queries with the api_key assigned to the `creator_role` or the master api key role. 

So we'd need an additional test: `with event trigger + ownership_role AND creator_role can DROP the tables`.

alrocar: Just realized the extension does not know anything about `creator_role` being a member of `ownership_role` so maybe we can just add this test to the Rails code, once everything is integrated.

ethervoid: We're going to test the removal using the creator role from rails tests because is there where we grant the ownership role membership to that role

ethervoid:
@alrocar final review with all :green_apple: ?

Makefile
```diff
@@ -171,7 +173,11 @@ legacy_regress: $(REGRESS_OLD) Makefile
 			$(SED) -e 's/@@VERSION@@/$(EXTVERSION)/' -e 's/@extschema@/cartodb/g' -e "s/@postgisschema@/public/g" >> $${of}; \
 		exp=expected/test/$${tn}.out; \
 		echo '\set ECHO none' > $${exp}; \
-		cat test/$${tn}_expect >> $${exp}; \
+		if [[ -f "test/$${tn}_expect.pg$(PG_VERSION)" ]]; then \
+			cat test/$${tn}_expect.pg$(PG_VERSION) >> $${exp}; \
+		else \
+			cat test/$${tn}_expect >> $${exp}; \
+		fi \
```
alrocar: 👍

gonzaloriestra:
Maybe I'm late to the party, but have you considered changing the tests a bit to avoid having different expectations depending on the PG version? That way we could also avoid adding more logic to the Makefile.

If I'm not wrong, the only difference between the expectations is the text of the `permission denied` error. What about if we mute the error output there and check that the action couldn't be completed instead (like no table created)?

ethervoid:
Welcome back!

Not late at all! From my PoV this is going to happen in more cases because of the differences between PG11/PG12...and the change in the Makefile is minimal so I think is a good solution (obviously I think/made it) but I see you point so I want to know what @alrocar thinks about it ;)

gonzaloriestra:
Yes, maybe we will need it in the future and it's a nice solution, but I'm just wondering if we could keep it simpler for now. As a con, the tests would be less explicit with the workaround.

alrocar:
For now I prefer being explicit in the expectations, thanks to that just by reading the expectations we can realize there are actually differences depending on the PG version, that otherwise would be hidden to us.

But yes this has room for improvement, like for example support regular expressions in the ERROR expectations. If you ask me, I won't do it by now... 👼

PR #369, opened 2019-07-12
User defined FDW's

rafatower:
This adds a couple of functions to ease set up of user foreign servers, mappings and tables.

Sample usage:
```sql
-- Create a FDW called "amazon"
SELECT cartodb.CDB_SetUp_User_Foreign_Server('amazon', '{
   "server": {
     "extensions": "postgis",
     "dbname": "testdb",
     "host": "localhost",
     "port": "5433"
   },
   "user_mapping": {
     "user": "fdw_user",
     "password": "secret"
   }
}');

-- Import a foreign table from "amazon"
SELECT cartodb.CDB_SetUp_User_Foreign_Table('amazon', 'carto-lite', 'cliwoc21');
```

NOTE: it is decoupled from carto-lite

It creates a role with the same name of the fdw (in the example, "amazon") and grants it to the caller of the function. So, by default just the caller can access the server, mappings and anything related to it, but they can also grant it to other users (including `publicuser`, the org role or even `public`).

E.g:

```sql
-- not the caller role
SELECT * FROM amazon.cliwoc21 limit 1;
ERROR:  permission denied for schema amazon
LINE 1: select * from amazon.cliwoc21 limit 1;

IMPORT FOREIGN SCHEMA carto_lite LIMIT TO (cliwoc21) FROM SERVER amazon INTO try_import;
ERROR:  permission denied for foreign server amazon
```

but we want to grant permissions to it:
```sql
GRANT amazon TO "development_cartodb_user_699ab2dc-dd77-4395-8c95-3a8c973dfafe" ;
```

and from that moment on:
```
SELECT current_user;
                         current_user                          
---------------------------------------------------------------
 development_cartodb_user_699ab2dc-dd77-4395-8c95-3a8c973dfafe
(1 row)

SELECT count(1) FROM amazon.cliwoc21 ;
 count  
--------
 282322
(1 row)
```

### DIFF
--- a/scripts-available/CDB_ForeignTable.sql
+++ b/scripts-available/CDB_ForeignTable.sql
@@ -139,6 +139,15 @@ $$
 LANGUAGE plpgsql VOLATILE PARALLEL UNSAFE;
 
 
+-- Produce a valid DB name for objects created for the user FDW's
+CREATE OR REPLACE FUNCTION @extschema@.__CDB_User_FDW_Object_Names(fdw_input_name NAME)
+RETURNS NAME AS $$
+  -- Note on input we use %s and on output we use %I, in order to
+  -- avoid double escaping
+  SELECT format('cdb_fdw_%s', fdw_input_name)::NAME;
+$$
+LANGUAGE sql IMMUTABLE PARALLEL SAFE;
+
 -- A function to set up a user-defined foreign data server
 -- It does not read from CDB_Conf.
 -- Only superuser roles can invoke it successfully
@@ -159,22 +168,23 @@ LANGUAGE plpgsql VOLATILE PARALLEL UNSAFE;
 --
 -- Underneath it will:
 --   * Set up postgresql_fdw
---   * Create a server with the name 'amazon'
---   * Create a role called 'amazon' to manage access
---   * Create a user mapping with that role 'amazon'
---   * Create a schema 'amazon' as a convenience to set up all foreign
+--   * Create a server with the name 'cdb_fdw_amazon'
+--   * Create a role called 'cdb_fdw_amazon' to manage access
+--   * Create a user mapping with that role 'cdb_fdw_amazon'
+--   * Create a schema 'cdb_fdw_amazon' as a convenience to set up all foreign
 --     tables over there
 --
 -- It is the responsibility of the superuser to grant that role to either:
 --   * Nobody
 --   * Specific roles: GRANT amazon TO role_name;
---   * Members of the organization: SELECT cartodb.CDB_Organization_Grant_Role('amazon');
---   * The publicuser: GRANT amazon TO publicuser;
-CREATE OR REPLACE FUNCTION @extschema@._CDB_SetUp_User_PG_FDW_Server(fdw_name NAME, config json)
+--   * Members of the organization: SELECT cartodb.CDB_Organization_Grant_Role('cdb_fdw_amazon');
+--   * The publicuser: GRANT cdb_fdw_amazon TO publicuser;
+CREATE OR REPLACE FUNCTION @extschema@._CDB_SetUp_User_PG_FDW_Server(fdw_input_name NAME, config json)
 RETURNS void AS $$
 DECLARE
   row record;
   option record;
+  fdw_objects_name NAME := @extschema@.__CDB_User_FDW_Object_Names(fdw_input_name);
 BEGIN
   -- TODO: refactor with original function
   -- This function tries to be as idempotent as possible, by not creating anything more than once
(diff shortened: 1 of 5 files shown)


rafatower:
@dgaubert to take an early look. Not final, cause it lacks some changes, a couple utility functions and documentation. Hopefully it'll help us decouple set up from other tasks (e.g: testing through SQL API)

rafatower:
For the record, `SECURITY DEFINER` cannot be really removed: https://github.com/CartoDB/cartodb-postgresql/commit/67663c79aa48cab8f3d8dd87b7a2a62a157bf0c3 (at least not in an easy and clean way)

scripts-available/CDB_ForeignTable.sql
```diff
@@ -139,6 +139,123 @@ $$
...
+
+
+-- Set up a user foreign table
+-- E.g:
+--   SELECT cartodb.CDB_SetUp_User_Foreign_Table('amazon', 'carto_lite', 'mytable');
+--   SELECT * FROM amazon.my_table;
+CREATE OR REPLACE FUNCTION @extschema@.CDB_SetUp_User_Foreign_Table(fdw_name NAME, foreign_schema NAME, table_name NAME)
+RETURNS void AS $$
+BEGIN
+  EXECUTE FORMAT ('IMPORT FOREIGN SCHEMA %I LIMIT TO (%I) FROM SERVER %I INTO %I;', foreign_schema, table_name, fdw_name, fdw_name);
```
dgaubert: Do we need to import the entire schema every time we want to add a foreign table? Or is it smart enough to detect only the missing tables?

scripts-available/CDB_ForeignTable.sql
```diff
@@ -139,6 +139,123 @@ $$
...
+--   * Create a user mapping with that role 'amazon'
+--   * Create a schema 'amazon' as a convenience to set up all foreign
+--     tables over there
+--
+-- It is the responsibility of the caller to grant that role to either:
+--   * Nobody
+--   * Specific roles: GRANT amazon TO role_name;
+--   * Members of the organization: SELECT cartodb.CDB_Grant_Role_To_Org_Members('amazon'); TODO
+--   * The publicuser: GRANT amazon TO publicuser;
+CREATE OR REPLACE FUNCTION @extschema@.CDB_SetUp_User_Foreign_Server(fdw_name NAME, config json)
```
dgaubert: Do you think it's worth to have different names for `schema/user/server`? I mean, from the point of view of management having different objects with the very same name cloud mislead, couldn't it?

scripts-available/CDB_ForeignTable.sql
rafatower: Actually it uses `LIMIT TO (%I)` which only scans the table mentioned and as per PG docs is the preferred way. Otherwise you need to provide a mapping between columns and types, which is more error prone:

https://www.postgresql.org/docs/11/postgres-fdw.html#id-1.11.7.42.16

> It's essential that the data types and other properties of the columns declared in CREATE FOREIGN TABLE match the actual remote table. Column names must match as well, unless you attach column_name options to the individual columns to show how they are named in the remote table. In many cases, use of IMPORT FOREIGN SCHEMA is preferable to constructing foreign table definitions manually.

dgaubert: Pretty clever, thanks for sharing!

scripts-available/CDB_ForeignTable.sql
rafatower: I applied "convention over configuration" which IMO is the only way to simplify things. The alternative is what we did with the "operations manual", which is to expose the whole PG SQL syntax and require a superadmin to setup things.

They have the same name, but they can be listed in different views and are in different tables in the PG catalog, so I don't really see drawbacks with this approach.

That said, there's some middle ground which is to [prefix/suffix with types](https://softwareengineering.stackexchange.com/questions/14789/what-is-the-appeal-of-systems-hungarian) as the "convention", but I wouldn't do it as we don't really need to carry domain-type info with the names of different entities in PG.

scripts-available/CDB_ForeignTable.sql
```diff
@@ -139,6 +139,132 @@ $$
...
+    -- Grant the fdw role to the caller, and permissions to grant it to others
+    EXECUTE FORMAT ('GRANT %I TO %I WITH ADMIN OPTION', fdw_name, session_user);
+
+    -- TODO: Bring here the remote cdb_tablemetadata
+END
+$$ LANGUAGE plpgsql VOLATILE PARALLEL UNSAFE SECURITY DEFINER;
+
+CREATE OR REPLACE FUNCTION @extschema@.CDB_Drop_User_Foreign_Server(fdw_name NAME)
+RETURNS void AS $$
+BEGIN
```
dgaubert: Should we drop the role as well?

rafatower: sure! I forgot it, thanks!

rafatower: It may be a little bit troublesome:

https://www.postgresql.org/docs/11/sql-droprole.html
> to drop non-superuser roles, you must have CREATEROLE privilege

https://www.postgresql.org/docs/11/sql-createrole.html

> Be careful with the CREATEROLE privilege. There is no concept of inheritance for the privileges of a CREATEROLE-role. That means that even if a role does not have a certain privilege but is allowed to create other roles, it can easily create another role with different privileges than its own (except for creating roles with superuser privileges). For example, if the role “user” has the CREATEROLE privilege but not the CREATEDB privilege, nonetheless it can create a new role with the CREATEDB privilege. Therefore, regard roles that have the CREATEROLE privilege as almost-superuser-roles.

So I might have to resort to some form of `SECURITY DEFINER` again. Let me check it...

rafatower: Addressed here: https://github.com/CartoDB/cartodb-postgresql/pull/369/commits/99096d41e0bbd65f77f798fade32e0353a235365

rafatower:
I think this is ready for a review. It lacks documentation but I'd rather write and "beta test" it through the "operations manual" than document here yet. No strong opinion, though.

scripts-available/CDB_ForeignTable.sql
dgaubert: 👌

scripts-available/CDB_ForeignTable.sql
```diff
@@ -139,6 +139,142 @@ $$
...
+
+    -- Give the fdw role ownership over the schema
+    EXECUTE FORMAT ('ALTER SCHEMA %I OWNER TO %I', fdw_name, fdw_name);
+
+    -- Grant the fdw role to the caller, and permissions to grant it to others
+    EXECUTE FORMAT ('GRANT %I TO %I WITH ADMIN OPTION', fdw_name, session_user);
+
+    -- TODO: Bring here the remote cdb_tablemetadata
+END
+$$ LANGUAGE plpgsql VOLATILE PARALLEL UNSAFE SECURITY DEFINER;
```
Algunenano: Do we really want to allow any user (specially those within organizations) to call this and create:
- New users.
- New schemas.
- New grants.

That looks dangerous and easily exploitable.

scripts-available/CDB_ForeignTable.sql
```diff
@@ -139,6 +139,142 @@ $$
...
+--   * Create a user mapping with that role 'amazon'
+--   * Create a schema 'amazon' as a convenience to set up all foreign
+--     tables over there
+--
+-- It is the responsibility of the caller to grant that role to either:
+--   * Nobody
+--   * Specific roles: GRANT amazon TO role_name;
+--   * Members of the organization: SELECT cartodb.CDB_Organization_Grant_Role('amazon');
+--   * The publicuser: GRANT amazon TO publicuser;
+CREATE OR REPLACE FUNCTION @extschema@.CDB_SetUp_User_Foreign_Server(fdw_name NAME, config json)
```
Algunenano: Since the function doesn't have a way to select the extension (`postgres_fdw` vs `odbc_fdw`), I'd be more explicit in the name of the function to make it clear this is for a remote PG server.

scripts-available/CDB_ForeignTable.sql
```diff
@@ -139,6 +139,142 @@ $$
...
+        THEN
+        EXECUTE FORMAT('ALTER SERVER %I OPTIONS (ADD %I %L)', fdw_name, row.key, row.value);
+      ELSE
+        EXECUTE FORMAT('ALTER SERVER %I OPTIONS (SET %I %L)', fdw_name, row.key, row.value);
+      END IF;
+    END LOOP;
+
+    -- Create specific role for this
+    IF NOT EXISTS ( SELECT 1 FROM pg_roles WHERE rolname = fdw_name) THEN
+       EXECUTE format('CREATE ROLE %I NOLOGIN', fdw_name);
```
Algunenano: I'm not sure if I like blindly using the name that was given as input. Can we add some prefix so we now in the future where it comes from? Something like `fdw_$username` or something similar.

scripts-available/CDB_ForeignTable.sql
```diff
@@ -139,6 +139,142 @@ $$
...
+      EXECUTE FORMAT ('CREATE SCHEMA %I', fdw_name);
+    END IF;
+
+    -- Give the fdw role ownership over the schema
+    EXECUTE FORMAT ('ALTER SCHEMA %I OWNER TO %I', fdw_name, fdw_name);
+
+    -- Grant the fdw role to the caller, and permissions to grant it to others
+    EXECUTE FORMAT ('GRANT %I TO %I WITH ADMIN OPTION', fdw_name, session_user);
+
+    -- TODO: Bring here the remote cdb_tablemetadata
```
Algunenano: I think we need to print notices with the information (`remote server X created`, `role Y created`) either inline with the changes or at the end.

scripts-available/CDB_ForeignTable.sql
```diff
@@ -139,6 +139,142 @@ $$
...
+        END IF;
+    END LOOP;
+
+    -- Grant usage on the wrapper and server to the fdw role
+    EXECUTE FORMAT ('GRANT USAGE ON FOREIGN DATA WRAPPER postgres_fdw TO %I', fdw_name);
+    EXECUTE FORMAT ('GRANT USAGE ON FOREIGN SERVER %I TO %I', fdw_name, fdw_name);
+
+    -- Create schema if it does not exist.
+    IF NOT EXISTS ( SELECT * from pg_namespace WHERE nspname=fdw_name) THEN
+      EXECUTE FORMAT ('CREATE SCHEMA %I', fdw_name);
```
Algunenano: Several projects from backend wrongly assume that the user can't create new schemas, so we should make a note about this and document it clearly, since this might cause issues with migrations / user mover.

scripts-available/CDB_ForeignTable.sql
```diff
@@ -139,6 +139,142 @@ $$
...
+CREATE OR REPLACE FUNCTION @extschema@.CDB_Drop_User_Foreign_Server(fdw_name NAME)
+RETURNS void AS $$
+BEGIN
+    EXECUTE FORMAT ('DROP SCHEMA %I', fdw_name);
+    EXECUTE FORMAT ('DROP USER MAPPING FOR public SERVER %I', fdw_name);
+    EXECUTE FORMAT ('DROP SERVER %I', fdw_name);
+    EXECUTE FORMAT ('REVOKE USAGE ON FOREIGN DATA WRAPPER postgres_fdw FROM %I', fdw_name);
+    EXECUTE FORMAT ('DROP ROLE %I', fdw_name);
+END
+$$ LANGUAGE plpgsql VOLATILE PARALLEL UNSAFE SECURITY DEFINER;
```
Algunenano: Same as above, do we want any user to be able to drop all this?

scripts-available/CDB_ForeignTable.sql
rafatower: Yes, you have a point there security-wise.

The alternatives I see is to either give permissions to it to specific users (breaking a bit `cartodb` design) or to let only superuser define those, in which case we lose the ability to grant to the caller.

Do you see a better option?

scripts-available/CDB_ForeignTable.sql
rafatower: Okay

scripts-available/CDB_ForeignTable.sql
rafatower: Okay

scripts-available/CDB_ForeignTable.sql
rafatower: if that helps... it's quick and easy to do :)

scripts-available/CDB_ForeignTable.sql
rafatower: How about `cdb_pgfdw_name` for those DB objects? server, schema, role.

scripts-available/CDB_ForeignTable.sql
Algunenano: No, I don't see a better option.

I think that the safe route is to have `CDB_SetUp_User_Foreign_Server` and `CDB_Drop_User_Foreign_Server` only working for superusers, then and extra function to make it easier granting priviledges to a different user to certain server (access to all tables in the new schema I guess), and finally add an extra option in `CDB_Drop_User_Foreign_Server` to force drop in cascade so it's easier to remote everything (set to false by default).

scripts-available/CDB_ForeignTable.sql
rafatower: Thanks for pointing that out. Do you know by any chance where it is implemented (anywhere other than `REVOKE ALL`) and where it is documented? this kind of "implicit limitation" is what makes direct connections and stuff like that hard to assess.

scripts-available/CDB_ForeignTable.sql
Algunenano: I'd say `cdb_fdw_name` is enough since in this case we don't need to be explicit about postgres_fdw vs other extensions. Also, I wouldn't change the name since that's an input parameter, just the objects that we create based on that.

scripts-available/CDB_ForeignTable.sql
```diff
@@ -139,6 +139,15 @@ $$
 LANGUAGE plpgsql VOLATILE PARALLEL UNSAFE;
 
 
+-- Produce a valid DB name for objects created for the user FDW's
+CREATE OR REPLACE FUNCTION @extschema@.__CDB_User_FDW_Object_Names(fdw_input_name NAME)
+RETURNS NAME AS $$
+  -- Note on input we use %s and on output we use %I, in order to
+  -- avoid double escaping
+  SELECT format('cdb_fdw_%s', fdw_input_name)::NAME;
```
Algunenano: Should we enclose this also under `quote_ident` to prevent odd things like spaces and possible injections (although the user already has full access, so them being able to inject code shouldn't be a concern :D).

rafatower: I'm applying the principle of "sanitize on input, escaping on output" (couldn't find a good reference to it on the internets, sorry). I also added an intentional comment about that in the code (BTW, `%I` is equivalent to `quote_ident`).

Let me explain it a little further:
- The function takes a `NAME` type, which is meant to host identifiers, but perhaps requires quoting such as in this `valid-name-with-dashes` which has been always a nuance. It could be tempting to use more specific types (and it is always a good idea, when possible), but cannot in this case as those objects do not exist yet. E.g:
```
postgres=# SELECT 'hola'::regnamespace;
ERROR:  schema "hola" does not exist
```

- Wherever I use the generated name, I use `format('%I')` to "escape on output".

- I modified the tests to use dashes in the base name, just to check there's no missing escaping or double escaping.

For the sake of the experiment, let's follow it with a malicious input:

```sql
-- Ok, evil input
SELECT cartodb.__CDB_User_FDW_Object_Names('evil_sufix; DROP TABLE users');
     __cdb_user_fdw_object_names      
--------------------------------------
 cdb_fdw_evil_sufix; DROP TABLE users
(1 row)

-- But escaping "on output"
SELECT format('CREATE SERVER %I FOREIGN DATA WRAPPER postgres_fdw', cartodb.__CDB_User_FDW_Object_Names('evil_sufix; DROP TABLE users'));
                                         format                                         
----------------------------------------------------------------------------------------
 CREATE SERVER "cdb_fdw_evil_sufix; DROP TABLE users" FOREIGN DATA WRAPPER postgres_fdw
(1 row)

-- As horrible as it is, it generated a valid identifier that requires quoting
-- We can run the query...
CREATE SERVER "cdb_fdw_evil_sufix; DROP TABLE users" FOREIGN DATA WRAPPER postgres_fdw;
CREATE SERVER

-- and finally delete it
DROP SERVER "cdb_fdw_evil_sufix; DROP TABLE users";
DROP SERVER
```

Mind that this will require superuser privileges to execute, or it will fail at the first step and the whole transaction (function call or enclosing transaction) will be rolled back.

If you still deem it necessary, the only thing I can think of is to get stricter (and custom) sanitization on input. If you could think of other approaches (not just for this piece of code but more general to the extensions) I'm all ears.

Algunenano: > BTW, %I is equivalent to quote_ident

Nice, I didn't know. I'd say that's enough, as going custom for validation is bound to give us a lot of headaches.

PR #372, opened 2019-10-08
PG federated server setup

rafatower:
This branch adds a new function `CDB_SetUp_PG_Federated_Server` that can be used like this:
```sql
SELECT cartodb.CDB_SetUp_PG_Federated_Server('amazon', '{
   "server": {
     "dbname": "testdb",
     "host": "myhostname.us-east-2.rds.amazonaws.com",
     "port": "5432"
   },
   "credentials": {
     "username": "read_only_user",
     "password": "secret"
   }
}');
```

As a bonus it has a small refactor of tests, extracting some common setup and teardown from FDW tests.

PR #373, opened 2019-10-09
PG Federated Tables

rafatower:
Built on top of https://github.com/CartoDB/cartodb-postgresql/pull/372

This adds a function `CDB_SetUp_PG_Federated_Table` with the following form:

```sql
SELECT cartodb.CDB_SetUp_PG_Federated_Table(
  'amazon', -- mandatory, name of the federated server
  'schema_name', -- mandatory
  'table_name', -- mandatory
  'id_column_name', -- mandatory
  'geom_column_name', -- optional, preferably in 4326
  'webmercator_column_name', -- optional, must be in 3857 if present
);
```

### DIFF
--- a/scripts-available/CDB_PG_Federated_Tables.sql
+++ b/scripts-available/CDB_PG_Federated_Tables.sql
@@ -114,7 +114,7 @@ $$ LANGUAGE SQL;
 --    }
 -- }');
 --
-CREATE OR REPLACE FUNCTION @extschema@.CDB_SetUp_PG_Federated_Server(server_alias text, server_config jsonb)
+CREATE OR REPLACE FUNCTION @extschema@.CDB_SetUp_PG_Federated_Server(server_alias name, server_config jsonb)
 RETURNS void
 AS $$
 DECLARE
@@ -149,7 +149,7 @@ LANGUAGE PLPGSQL VOLATILE PARALLEL UNSAFE;
 -- );
 --
 CREATE OR REPLACE FUNCTION @extschema@.CDB_SetUp_PG_Federated_Table(
-    server_alias text,
+    server_alias name,
     schema_name name,
     table_name name,
     id_column name,
@@ -229,7 +229,7 @@ LANGUAGE PLPGSQL VOLATILE PARALLEL UNSAFE;
 
 
 CREATE OR REPLACE FUNCTION @extschema@.CDB_SetUp_PG_Federated_Table(
-    server_alias text,
+    server_alias name,
     schema_name name,
     table_name name,
     id_column name,
@@ -250,7 +250,7 @@ LANGUAGE SQL VOLATILE PARALLEL UNSAFE;
 
 
 CREATE OR REPLACE FUNCTION @extschema@.CDB_SetUp_PG_Federated_Table(
-    server_alias text,
+    server_alias name,
     schema_name name,
     table_name name,
     id_column name
--- a/scripts-available/CDB_PG_Federated_Tables.sql
+++ b/scripts-available/CDB_PG_Federated_Tables.sql
@@ -45,6 +45,7 @@ LANGUAGE PLPGSQL IMMUTABLE PARALLEL SAFE;
 
 --
 -- Set up a federated server for later connection of tables/views
+--
 -- E.g:
 -- SELECT cartodb.CDB_SetUp_PG_Federated_Server('amazon', '{
 --    "server": {
@@ -67,6 +68,48 @@ BEGIN
         @extschema@.__ft_add_default_options(server_config)
     );
     PERFORM cartodb._CDB_SetUp_User_PG_FDW_Server(server_alias, final_config::json);
+END
+$$
+LANGUAGE PLPGSQL VOLATILE PARALLEL UNSAFE;
+
+
+--
+-- Set up a federated table
+--
+-- E.g:
+-- SELECT cartodb.CDB_SetUp_PG_Federated_Table(
+--   'amazon',                  -- mandatory, name of the federated server
+--   'my_remote_schema',        -- mandatory, schema name
+--   'my_remote_table',         -- mandatory, table name
+--   'id',                      -- mandatory, name of the id column
+--   'geom',                    -- optional, name of the geom column, preferably in 4326
+--   'webmercator_column_name', -- optional, must be in 3857 if present
+-- );
+CREATE OR REPLACE FUNCTION @extschema@.CDB_SetUp_PG_Federated_Table(
+    server_alias text,
+    schema_name name,
+    table_name name,
+    id_colum_name name,
+    geom_column_name name,
+    webmercator_column_name name
+)
+RETURNS void
+AS $$
+DECLARE
+    fdw_objects_name NAME := @extschema@.__CDB_User_FDW_Object_Names(server_alias);
+BEGIN
+    -- Get the table
+    PERFORM CDB_SetUp_User_PG_FDW_Table(server_alias, schema_name, table_name);
+
+    -- Create the view
+    EXECUTE format(
+        'CREATE OR REPLACE VIEW %1$I AS SELECT * FROM %2$I.%1$s',
+        table_name,
+        fdw_objects_name
+    );
+
+    -- Grant perms to the view
+    EXECUTE format('GRANT SELECT ON %I TO %s', table_name, fdw_objects_name);
 END
 $$
 LANGUAGE PLPGSQL VOLATILE PARALLEL UNSAFE;
--- a/scripts-available/CDB_PG_Federated_Tables.sql
+++ b/scripts-available/CDB_PG_Federated_Tables.sql
@@ -43,8 +43,8 @@ $$
 LANGUAGE PLPGSQL IMMUTABLE PARALLEL SAFE;
 
 
-CREATE OR REPLACE FUNCTION @extschema@.__ft_assert_numeric(input_table regclass, colname name)
-RETURNS VOID
+CREATE OR REPLACE FUNCTION @extschema@.__ft_is_numeric(input_table regclass, colname name)
+RETURNS boolean
 AS $$
 BEGIN
     PERFORM atttypid FROM pg_catalog.pg_attribute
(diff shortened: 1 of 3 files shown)


rafatower:
Tests are failing because of an issue with repositories: https://travis-ci.org/CartoDB/cartodb-postgresql/jobs/595709511#L473

scripts-available/CDB_PG_Federated_Tables.sql
```diff
@@ -0,0 +1,295 @@
+----------------------------------------------------------------------
+-- Federated Tables management functions
+----------------------------------------------------------------------
+
+-- Take a config jsonb and transform it to an input suitable for
+-- _CDB_SetUp_User_PG_FDW_Server
+CREATE OR REPLACE FUNCTION @extschema@.__ft_credentials_to_user_mapping(input_config jsonb)
```
Algunenano: What does ft mean?

test/extension/test.sh
```diff
@@ -651,17 +670,163 @@ EOF
...
+        "port": ${PGPORT:-5432}
+    },
+    "credentials": {
+        "username": "fdw_user",
+        "password": "foobarino"
+    }
 }
+EOF
+
+    # Unit-test __ft_credentials_to_user_mapping
```
Algunenano: `__ft_credentials_to_user_mapping` needs more tests to exemplify what's the expected input and output

scripts-available/CDB_PG_Federated_Tables.sql
Algunenano: Same question for the other functions. Whatever that is, is it necessary?

scripts-available/CDB_PG_Federated_Tables.sql
```diff
@@ -0,0 +1,295 @@
...
+
+
+-- Take a config jsonb as input and return it augmented with default
+-- options
+CREATE OR REPLACE FUNCTION @extschema@.__ft_add_default_options(input_config jsonb)
+RETURNS jsonb
+AS $$
+DECLARE
+    default_options jsonb := '{
+        "extensions": "postgis",
```
Algunenano: Do we want to have `postgis` on by default without checking the remote server? Should we leave it out until we can check it in a following iteration?

scripts-available/CDB_PG_Federated_Tables.sql
```diff
@@ -0,0 +1,295 @@
...
+        "updatable": "false",
```
Algunenano: It's really nice to add this option (so we don't depend on the configuration of the customer in their server), but I'd change the name of the function to reflect these are the default `read only` options.

scripts-available/CDB_PG_Federated_Tables.sql
```diff
@@ -0,0 +1,295 @@
...
+    RETURN jsonb_set(input_config, '{server}'::text[], server_config);
+END
+$$
+LANGUAGE PLPGSQL IMMUTABLE PARALLEL SAFE;
+
+
+CREATE OR REPLACE FUNCTION @extschema@.__ft_assert_numeric(input_table regclass, colname name)
+RETURNS VOID
+AS $$
+BEGIN
```
Algunenano: Do we need to check for NULLs too?

scripts-available/CDB_PG_Federated_Tables.sql
```diff
@@ -0,0 +1,295 @@
...
+    PERFORM atttypid FROM pg_catalog.pg_attribute
+       WHERE attrelid = input_table
+         AND attname = colname
+         AND atttypid IN (SELECT oid FROM pg_type
+           WHERE typname IN
+             ('smallint', 'integer', 'bigint', 'int2', 'int4', 'int8'));
+    IF NOT FOUND THEN
+      RAISE EXCEPTION 'non integer id_column "%"', colname;
```
Algunenano: I think it'd better to return true / false rather than return void and raise a costly exception.

scripts-available/CDB_PG_Federated_Tables.sql
```diff
@@ -0,0 +1,295 @@
...
+CREATE OR REPLACE FUNCTION @extschema@.__ft_assert_geometry(input_table regclass, colname name)
+RETURNS VOID
+AS $$
+BEGIN
+    PERFORM atttypid FROM pg_catalog.pg_attribute
+        WHERE attrelid = input_table
+           AND attname = colname
+           AND atttypid = 'geometry'::regtype;
+    IF NOT FOUND THEN
+        RAISE EXCEPTION 'non geometry column "%"', colname;
```
Algunenano: Same as above, I think returning bool is better since it allow us to act on it (and not only on failure).

scripts-available/CDB_PG_Federated_Tables.sql
```diff
@@ -0,0 +1,295 @@
...
+CREATE OR REPLACE FUNCTION @extschema@.CDB_SetUp_PG_Federated_Server(server_alias text, server_config jsonb)
+RETURNS void
+AS $$
+DECLARE
+    final_config jsonb;
+BEGIN
+    final_config := @extschema@.__ft_credentials_to_user_mapping(
+        @extschema@.__ft_add_default_options(server_config)
+    );
+    PERFORM cartodb._CDB_SetUp_User_PG_FDW_Server(server_alias, final_config::json);
```
Algunenano: `cartodb` here should be `@extschema@`

scripts-available/CDB_PG_Federated_Tables.sql
```diff
@@ -0,0 +1,295 @@
...
+AS $$
+DECLARE
+    fdw_objects_name NAME := @extschema@.__CDB_User_FDW_Object_Names(server_alias);
+    src_table REGCLASS;
+    rest_of_cols TEXT[];
+    geom_expression TEXT;
+    webmercator_expression TEXT;
+BEGIN
+    -- Import the foreign table
+    PERFORM CDB_SetUp_User_PG_FDW_Table(server_alias, schema_name, table_name);
```
Algunenano: Does this need `@extschema@`

scripts-available/CDB_PG_Federated_Tables.sql
```diff
@@ -0,0 +1,295 @@
...
+    PERFORM @extschema@.__ft_assert_geometry(src_table, webmercator_column);
+
+    -- Get a list of columns excluding the id, geom and the_geom_webmercator
+    SELECT ARRAY(
+        SELECT quote_ident(c) FROM @extschema@.__ft_getcolumns(src_table) AS c
+        WHERE c NOT IN (id_column, geom_column, webmercator_column)
+    ) INTO rest_of_cols;
+
+    -- Figure out whether a ST_Transform to 4326 is needed or not
+    IF Find_SRID(fdw_objects_name::varchar, table_name::varchar, geom_column::varchar) = 4326
```
Algunenano: This needs @postgisschema@, same bellow.

Algunenano: Same for `ST_Transform` calls,

scripts-available/CDB_PG_Federated_Tables.sql
```diff
@@ -0,0 +1,295 @@
...
+)
+RETURNS void
+AS $$
+DECLARE
+    fdw_objects_name NAME := @extschema@.__CDB_User_FDW_Object_Names(server_alias);
+    src_table REGCLASS;
+    rest_of_cols TEXT[];
+BEGIN
+    -- Import the foreign table
+    PERFORM CDB_SetUp_User_PG_FDW_Table(server_alias, schema_name, table_name);
```
Algunenano: Needs schema qualification

scripts-available/CDB_PG_Federated_Tables.sql
rafatower: `__ft_*` stands for "federated table" and I'm copying that pattern from good old C, as poor man's way of keeping private functions in their own "namespace"

scripts-available/CDB_PG_Federated_Tables.sql
rafatower: input params? why? I don't think so but may be wrong

scripts-available/CDB_PG_Federated_Tables.sql
rafatower: I'd raise the exception in the caller then But no strong opinion.

What I wouldn't do is to return a success true/false value from the "public" function.

scripts-available/CDB_PG_Federated_Tables.sql
rafatower: cachis

scripts-available/CDB_PG_Federated_Tables.sql
Algunenano: I mean that  the column contents might be NULL.

rafatower: Yes, and there may be duplicates, but there's little we can do about it without an "aggressive" check to inform the user.

I'd add that check only if we see it does save us some "support time".

scripts-available/CDB_PG_Federated_Tables.sql
rafatower: done

scripts-available/CDB_PG_Federated_Tables.sql
rafatower: I'll leave it "as is" as our initial target is still databases with postgis installed.

scripts-available/CDB_PG_Federated_Tables.sql
```diff
@@ -0,0 +1,299 @@
...
+    src_table REGCLASS;
+    rest_of_cols TEXT[];
+BEGIN
+    -- Import the foreign table
+    PERFORM @extschema@.CDB_SetUp_User_PG_FDW_Table(server_alias, schema_name, table_name);
+    src_table := format('%s.%s', fdw_objects_name, table_name);
+
+    -- Check id_column is numeric
+    IF NOT @extschema@.__ft_is_numeric(src_table, id_column) THEN
+        RAISE EXCEPTION 'non integer id_column "%"', id_colun;
```
Algunenano: The id should be `id_column`, as it is it's not outputting the correct name. This is something I'd expect the tests had covered, instead I was getting a super odd error message.

scripts-available/CDB_PG_Federated_Tables.sql
```diff
@@ -0,0 +1,299 @@
...
+    geom_expression TEXT;
+    webmercator_expression TEXT;
+BEGIN
+    -- Import the foreign table
+    PERFORM @extschema@.CDB_SetUp_User_PG_FDW_Table(server_alias, schema_name, table_name);
+    src_table := format('%s.%s', fdw_objects_name, table_name);
+
+    -- Check id_column is numeric
+    IF NOT @extschema@.__ft_is_numeric(src_table, id_column) THEN
+        RAISE EXCEPTION 'non integer id_column "%"', id_colun;
```
Algunenano: Same as above, it should be id_column.

Algunenano:
Superseeded / Included by https://github.com/CartoDB/cartodb-postgresql/pull/382

PR #374, opened 2019-10-15
Avoid name clashing no geom columns provided but existing the_geom

rafatower:
Fix for CDB_SetUp_PG_Federated_Table('server', 'schema', 'table', 'id)
when the table contains the_geom and/or the_geom_webmercator columns
but they are not part of the input.

Otherwise it fails with

    ERROR:  column "the_geom" specified more than once

### DIFF
--- a/scripts-available/CDB_PG_Federated_Tables.sql
+++ b/scripts-available/CDB_PG_Federated_Tables.sql
@@ -274,7 +274,7 @@ BEGIN
     -- Get a list of columns excluding the id
     SELECT ARRAY(
         SELECT quote_ident(c) FROM @extschema@.__ft_getcolumns(src_table) AS c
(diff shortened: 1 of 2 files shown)


rafatower:
let me include the missing case for providing just geom column...

Algunenano:
I think this is also going to clash if my remote table has `cartodb_id` but I don't use it as identifier.

rafatower:
The mandatory parameter is excluded as it is filtered by `WHERE c NOT IN (id_column...)`. I added it explicitly, for the sake of clarity.

Algunenano:
LGTM, but we should have tests for all these cases. I'm almost 100% confident we are going to modify these functions in the future to support more stuff, so these things should all be tested.

PR #377, opened 2019-10-21
PG Federated Tables: fix quoting of input NAME params

rafatower:
NAME params require quoting, as opposed to REGCLASS, as they are
automatically checked on input and quoted on output.

We cannot use REGCLASS on inputs cause the system catalog does not
contain the foreign objects yet.

OTOH we have to be careful not to quote twice.

E.g:

```sql
-- The bugfix
SELECT format('%s', 'my name'::NAME); -- => my name
SELECT format('%I', 'my name'::NAME); -- => "my name"

-- REGCLASS type checks the system catalog underneath
SELECT format('%s', 'this_does_not_exist'::regclass); -- => ERROR:  relation "this_does_not_exist" does not exist

-- Table names with spaces
CREATE TABLE "name with spaces"();
SELECT format('%s', 'name with spaces'::regclass);   -- => ERROR:  invalid name syntax
SELECT format('%s', '"name with spaces"'::regclass); -- => "name with spaces"

-- Double-quoting issues
SELECT format('%I', '"name with spaces"'::regclass); -- => """name with spaces"""
SELECT * FROM """name with spaces""";      -- => ERROR:  relation ""name with spaces"" does not exist
SELECT '"""name with spaces"""'::regclass; -- => ERROR:  relation ""name with spaces"" does not exist
```

### DIFF
--- a/test/extension/test.sh
+++ b/test/extension/test.sh
@@ -901,6 +901,20 @@ EOF
                           )"
     sql cdb_testmember_1 'SELECT * FROM "Remote table10";' should '1|||42'
 
+    # same as above with geom and mercator cols
+    DATABASE=fdw_target sql postgres 'CREATE TABLE test_fdw."Remote table11" ("my id" int, "my value" int, "my geom" geometry(Geometry,4326), "my mercator" geometry(Geometry,3857));'
+    DATABASE=fdw_target sql postgres 'INSERT INTO test_fdw."Remote table11" VALUES (1, 42, cartodb.CDB_LatLng(0,0));'
+    DATABASE=fdw_target sql postgres 'GRANT SELECT ON ALL TABLES IN SCHEMA test_fdw TO fdw_user;'
+    sql cdb_testmember_1 "SELECT cartodb.CDB_SetUp_PG_Federated_Table(
+                              'my_server',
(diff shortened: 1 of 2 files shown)


test/extension/test.sh
```diff
@@ -888,6 +888,19 @@ EOF
...
+    # It shall work with columns with especial characters (those that
+    # require quoting)
+    DATABASE=fdw_target sql postgres 'CREATE TABLE test_fdw."Remote table10" ("my id" int, "my value" int);'
+    DATABASE=fdw_target sql postgres 'INSERT INTO test_fdw."Remote table10" VALUES (1, 42);'
+    DATABASE=fdw_target sql postgres 'GRANT SELECT ON ALL TABLES IN SCHEMA test_fdw TO fdw_user;'
+    sql cdb_testmember_1 "SELECT cartodb.CDB_SetUp_PG_Federated_Table(
+                              'my_server',
+                              'test_fdw',
+                              'Remote table10',
+                              'my id'
```
Algunenano: Can you test also both the_geom and the_geom_webmercator columns, please?

rafatower: Added another test for it.

PR #378, opened 2019-10-23
PG 12 + PLPYTHON 3 compatibility

Algunenano:
- Adds PG12 to the matrix.
- Adds some missing qualifications.
- Since _CDB_Aggregable_Attributes and CDB_ColumnNames use `sql_identifier` which changed from `text` to `name`, it needs to be casted in several places so that it doesn't get truncated.
- QuotaTest no longer depends on random(). Pending hiding the kb numbers since they seem to change depending on the run (after regenerating the DB, appears in all PG versions).
- OAuth tests are no longer dependent on the PG version.
- Until https://redmine.postgresql.org/issues/4868 is clarified, this changes PG12 build to depend on plpython3 instead of plpython (which defaults to plpython2). Multiple functions had to be adapted to work both with python2 and python3.


scripts-available/CDB_Groups_API.sql
```diff
@@ -202,17 +223,17 @@ $$
     last_err = None
     while retry > 0:
       try:
-        client = SD['groups_api_client'] = httplib.HTTPConnection(params['host'], params['port'], False, params['timeout'])
+        conn = SD['groups_api_client'] = client.HTTPConnection(params['host'], params['port'], False, params['timeout'])
```
rafatower: Out of scope of the changes: I don't get how this is used. Seems to be copied from somewhere else, and I believe the original intent was to cache the connection in the [SD](https://www.postgresql.org/docs/current/plpython-sharing.html)

scripts-available/CDB_Quota.sql
```diff
@@ -20,33 +20,50 @@ RETURNS bigint AS
 $$
 DECLARE
   total_size INT8;
+  raster_available BOOLEAN;
+  raster_read_query TEXT;
 BEGIN
+  -- Postgis 3+ might not install raster
```
rafatower: We have to account for Postgis3 not including raster face to migrations. Although never officially supported, I wonder if there's someone depending on raster yet. cc/ @manmorjim

Algunenano:
> Let's make sure it works in our servers (covered by CI, staging for extra safety). At present I prefer your approach: being explicit about plpython3u.

Ups, wrong link. The correct one is https://redmine.postgresql.org/issues/4868. Right now I'm not too worried about what the decision would be, but it should be revisited when preparing the migration to PG12

scripts-available/CDB_Groups_API.sql
Algunenano: You are probably right. We can open a ticket to investigate later if you want.

scripts-available/CDB_Quota.sql
Algunenano: There were some cases that appeared during the last migration but, as of today, the idea is to drop postgis_raster and postgis_topology during the next migration, but we'll have to get into more detail about how.

Algunenano:
I'm merging this since it doesn't involve any API changes and it shouldn't mean any changes for PG10 or PG11, but it's a good base to iterate over.

PR #379, opened 2019-10-25
Fix OAuth ownership re-assignation for functions

gonzaloriestra:
Related to: https://github.com/CartoDB/cartodb/issues/15160

### DIFF
--- a/test/CDB_OAuth.sql
+++ b/test/CDB_OAuth.sql
@@ -2,7 +2,7 @@
 \set QUIET on
 SET client_min_messages TO error;
 
--- The permission error changed between pre PG11 and post 11 (before everythin "relation", now it's "view", "table" and so on
+-- The permission error changed between pre PG11 and post 11 (before everything was "relation", now it's "view", "table" and so on
 CREATE OR REPLACE FUNCTION catch_permission_error(query text)
     RETURNS bool
 AS $$
@@ -44,6 +44,8 @@ SELECT * FROM test_view;
 SELECT * FROM test_mview;
 SELECT * FROM test_selectinto;
 SELECT test_function();
+-- Postgres grants default execute privilege on functions to PUBLIC. So in order to check the different permissions
+-- between creator and owner roles is not enough with performing a selection, we need to DROP the table (which only the owner can do)
 DROP FUNCTION test_function();
 
 \set QUIET on
--- a/test/CDB_OAuth.sql
+++ b/test/CDB_OAuth.sql
@@ -36,12 +36,14 @@ CREATE TABLE test_tablesas AS SELECT * FROM test;
 CREATE VIEW test_view AS SELECT * FROM test;
 CREATE MATERIALIZED VIEW test_mview AS SELECT * FROM test;
 SELECT * INTO test_selectinto FROM test;
+CREATE FUNCTION test_function() RETURNS text AS $$ BEGIN RETURN 'test'; END; $$ LANGUAGE PLPGSQL;
 
 SELECT * FROM test;
 SELECT * FROM test_tablesas;
 SELECT * FROM test_view;
 SELECT * FROM test_mview;
 SELECT * FROM test_selectinto;
+SELECT test_function();
 
 \set QUIET on
 SET SESSION AUTHORIZATION "ownership_role";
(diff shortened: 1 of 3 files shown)


esloho:
Default privileges granted for functions are not the same as for other objects in postgres. As stated in the documentation:

> PostgreSQL grants default privileges on some types of objects to PUBLIC. No privileges are granted to PUBLIC by default on tables, table columns, sequences, foreign data wrappers, foreign servers, large objects, schemas, or tablespaces. For other types of objects, the default privileges granted to PUBLIC are as follows: CONNECT and TEMPORARY (create temporary tables) privileges for databases; EXECUTE privilege for functions; and USAGE privilege for languages and data types (including domains).

I updated the expected results of the tests to reflect this (the event trigger function can be executed publicly). However, if extra security is preferred we should explicitly limit the execute privileges of this trigger function to mimic the other objects' behavior (and change back the tests accordingly). @gonzaloriestra

gonzaloriestra:
@esloho Nice. I think there's no need to change the default behavior.

But now we are testing that it doesn't raise an error when executing and that it allows the ownership role to drop it, which is ok, but we should also check that other roles can't do it:

- Without trigger: only the creator role can drop.
- With trigger: only the ownership role can drop.

esloho:
> we should also check that other roles can't do it:
> 
> * Without trigger: only the creator role can drop.
> * With trigger: only the ownership role can drop.

Agree, uses cases added :)

test/CDB_OAuth.sql
```diff
@@ -36,14 +36,18 @@ CREATE TABLE test_tablesas AS SELECT * FROM test;
...
 SELECT * FROM test;
 SELECT * FROM test_tablesas;
 SELECT * FROM test_view;
 SELECT * FROM test_mview;
 SELECT * FROM test_selectinto;
+SELECT test_function();
+DROP FUNCTION test_function();
 
 \set QUIET on
+CREATE FUNCTION test_function() RETURNS integer AS $$ BEGIN RETURN 1; END; $$ LANGUAGE PLPGSQL;
```
Algunenano: Do you need to drop and recreate the function here?

Algunenano: Same with test2_function

test/CDB_OAuth.sql
```diff
@@ -52,6 +56,8 @@ SELECT 'denied_tableas', catch_permission_error($$SELECT * FROM test_tablesas;$$
 SELECT 'denied_view', catch_permission_error($$SELECT * FROM test_view;$$);
 SELECT 'denied_mview', catch_permission_error($$SELECT * FROM test_mview;$$);
 SELECT 'denied_selectinto', catch_permission_error($$SELECT * FROM test_selectinto;$$);
+SELECT test_function();
+SELECT 'denied_function', catch_permission_error($$DROP FUNCTION test_function();$$);
```
Algunenano: Could you leave a comment explaning why you are testing dropping the table instead of selecting it, plese? Something similar to https://github.com/CartoDB/cartodb-postgresql/pull/379#issuecomment-550487752

test/CDB_OAuth.sql
esloho: In order to cover the case in https://github.com/CartoDB/cartodb-postgresql/pull/379#issuecomment-550951903 and since we're testing that the owner can do what other users can't (like doing select for tables) we need to test that the drop can be done. In this step the action should be allowed and performed so we need to create the function again for the next check.

test/CDB_OAuth.sql
esloho: In these tests we want to ensure that only the owner can do "privileged" actions, and that the owner is set to the right user during the oAuth flow. In the case of functions, the only "privileged" action is drop, so in order to test this behavior the steps are:

- create the function from the `creator_role` session
- perform public action (select) which should be allowed
- perform privileged action (drop) which should be allowed
- create the function again
- change session to `owner_role`
- perform public action (select) which should be allowed
- perform privileged action (drop) --> result will depend on the use case being tested (the trigger being enabled or not plus the role having the necessary permissions)

In the last test use case, the first drop attempted from the `creator_role` session is expected to fail since the trigger is enabled and well configured and the owner is then rightly set to `owner_role` and only the owner can do the drop.

Algunenano: What I meant is, can you write it down with the tests so we know about it when we read it months from now?

esloho: Oh! Sorry I misunderstood it. Of course!

PR #380, opened 2019-10-28
Regress target

Algunenano:
Fixes the unpackaged generation and uses it to load the extension without requiring a install, so you can do `make regress` locally and it should run with the latest code and tests without modifying the system files.


Algunenano:
Thanks!.

I'm going to make some extra changes to allow passing PGPORT and PGUSER to the scripts so it's handled automagically (so we can test fdw in the same way as it is right now), but I'll add it in the PR with the SQL functions.

PR #381, opened 2019-10-29
PG federated tables listing

rafatower:
This implements listing of remote schemas and tables, from a federated server, registered or not.

A few important remarks:
- It does not depend on any extra extension (e.g `dblink` nor patches in `postgres_fdw`)
- It imports a couple foreign views into the local schema given by `__CDB_User_FDW_Object_Names()`. The target schema may be changed to avoid conflicts.
- Potentially they may return more info, such as the the [`table_type`](https://www.postgresql.org/docs/11/infoschema-tables.html). I stuck to the essence until we clear out some doubts
- IMO there should be functions to list registered tables and to list ALL foreign tables (registered or not). API-wise this may be solved by adding a column `registered bool` IF we have a solid convention to place federated objects and at the expense of complexity.

@Algunenano at this point I'm more concerned about the interface and the relations between the functions than in the implementation. What do you think?

Thanks!

### DIFF
--- a/test/CDB_Federate_Tables_Listing_Test.sql
+++ b/test/CDB_Federate_Tables_Listing_Test.sql
@@ -1,0 +1,89 @@
+-- ===================================================================
+-- create FDW objects
+-- ===================================================================
+\set QUIET on
+CREATE EXTENSION IF NOT EXISTS postgres_fdw;
+
+CREATE SERVER testserver1 FOREIGN DATA WRAPPER postgres_fdw;
+DO $d$
+    BEGIN
+        EXECUTE $$CREATE SERVER loopback FOREIGN DATA WRAPPER postgres_fdw
+            OPTIONS (dbname '$$||current_database()||$$',
+                     port '$$||current_setting('port')||$$'
+            )$$;
+        EXECUTE $$CREATE SERVER loopback2 FOREIGN DATA WRAPPER postgres_fdw
+            OPTIONS (dbname '$$||current_database()||$$',
+                     port '$$||current_setting('port')||$$'
+            )$$;
+    END;
+$d$;
+
+CREATE USER MAPPING FOR CURRENT_USER SERVER loopback;
+CREATE USER MAPPING FOR CURRENT_USER SERVER loopback2;
+
+-- ===================================================================
+-- create objects used through FDW loopback server
+-- ===================================================================
+CREATE TYPE user_enum AS ENUM ('foo', 'bar', 'buz');
+CREATE SCHEMA "S 1";
+CREATE TABLE "S 1"."T 1" (
+	"C 1" int NOT NULL,
+	c2 int NOT NULL,
+	c3 text,
+	c4 timestamptz,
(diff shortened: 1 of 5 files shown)


test/CDB_Federate_Tables_Listing_Test_expect
```diff
@@ -0,0 +1,14 @@
...
+public
+Test listing of schemas from an unsupported server
+ERROR:  Not implemented server type file_fdw for remote server pglog
+Test listing of remote tables (sunny day)
+T 1
+T 2
+T 3
+T 4
+Test listing of remote tables from an unsupported server
+ERROR:  function cartodb.cdb_federated_server_list_remote_schemas(remote_server => unknown, remote_schema => unknown) does not exist at character 15
```
Algunenano: This isn't the error we are looking for.

rafatower: Good catch. I've been messing a little with different interfaces before pushing. Fix in progress...

Algunenano:
> It does not depend on any extra extension (e.g dblink nor patches in postgres_fdw)
This is great.

> @Algunenano at this point I'm more concerned about the interface and the relations between the functions than in the implementation. What do you think?

Agreed. This is going to break against the conventions I'm using for the other functions. Do you mind if I take these functions and incorporate them into my branch / PR in a consistent way?

> IMO there should be functions to list registered tables and to list ALL foreign tables (registered or not). API-wise this may be solved by adding a column registered bool IF we have a solid convention to place federated objects and at the expense of complexity.

I'm going to keep them separated for now, we can join them before merge / release.

rafatower:
> Do you mind if I take these functions and incorporate them into my branch / PR in a consistent way?

Not at all! On the contrary, I really appreciate it :heart:  Glad we're not stepping too much on each other's toes too.

test/CDB_Federate_Tables_Listing_Test.sql
```diff
@@ -0,0 +1,106 @@
...
+ALTER TABLE "S 1"."T 4" SET (autovacuum_enabled = 'false');
+\set QUIET off
+
+
+-- ===================================================================
+-- Test the listing functions
+-- ===================================================================
+\echo 'Test listing of remote schemas (sunny day)'
+SELECT * FROM cartodb.CDB_Federated_Server_List_Remote_Schemas(remote_server => 'loopback')
+    WHERE remote_schema NOT LIKE 'pg_%' -- Exclude toast and temp schemas
```
Algunenano: @rafatower Do you think we should ignore all `pg_% schemas automatically in cartodb.CDB_Federated_Server_List_Remote_Schemas? I think i'd make sense

rafatower: Probably, yes. I cannot think of a case in which the user would be interested in registering PG internal tables.

(side note: the test setup is copied from postgres_fdw tests)

rafatower:
Closing in favor of Raul's branch

PR #383, opened 2019-11-08
Retrieve target PG and postgis versions

rafatower:
Using the proposal from Algunenano, I added just a public function that is meant to be augmented with more info.

Here the interface
```sql
CDB_Federated_Server_Diagnostics(internal_server name) returns jsonb
```

### DIFF
--- a/scripts-available/CDB_FederatedServerDiagnostics.sql
+++ b/scripts-available/CDB_FederatedServerDiagnostics.sql
@@ -93,8 +93,6 @@ DECLARE
 BEGIN
     -- See https://www.postgresql.org/docs/current/catalog-pg-foreign-server.html
     -- See https://www.postgresql.org/docs/current/functions-info.html
-
-    -- SELECT jsonb_build_object(opt.option_name, opt.option_value) FROM (SELECT (pg_options_to_table(srvoptions)).* FROM pg_foreign_server WHERE srvname = 'cdb_fdw_local_pg11') opt;
     FOR opt IN
         SELECT (pg_options_to_table(srvoptions)).* FROM pg_foreign_server WHERE srvname = server_internal
     LOOP
--- a/scripts-available/CDB_FederatedServerDiagnostics.sql
+++ b/scripts-available/CDB_FederatedServerDiagnostics.sql
@@ -13,7 +13,6 @@ DECLARE
     remote_schema name := 'pg_catalog';
     remote_table name := 'pg_settings';
     local_schema name := @extschema@.__CDB_FS_Create_Schema(server_internal, remote_schema);
-    role_name name := @extschema@.__CDB_FS_Generate_Server_Role_Name(server_internal);
     remote_server_version text;
 BEGIN
     -- Import the foreign pg_settings table
@@ -52,7 +51,6 @@ DECLARE
     remote_schema name := 'pg_catalog';
     remote_table name := 'pg_extension';
     local_schema name := @extschema@.__CDB_FS_Create_Schema(server_internal, remote_schema);
-    role_name name := @extschema@.__CDB_FS_Generate_Server_Role_Name(server_internal);
     remote_postgis_version text;
 BEGIN
     -- Import the foreign pg_extension table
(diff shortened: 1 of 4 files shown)


rafatower:
Example usage (not included in the tests, as they can return json keys in different order):
```sql
SELECT cartodb.CDB_Federated_Server_Diagnostics(server => 'loopback');
            cdb_federated_server_diagnostics            
--------------------------------------------------------
 {"server_version": "11.3", "postgis_version": "2.5.1"}
(1 row)
```

rafatower:
Thanks! I'll add the foreign server options before closing this, but wanna do the latency stuff in the separate ticket.

rafatower:
Added `server_options` to the mix:

```sql
SELECT cartodb.CDB_Federated_Server_Diagnostics(server => 'loopback');
                                                                                          cdb_federated_server_diagnostics                                                                                           
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 {"server_options": {"host": "localhost", "port": "5432", "updatable": "false", "extensions": "postgis", "fetch_size": "1000", "use_remote_estimate": "true"}, "server_version": "11.3", "postgis_version": "2.5.1"}
(1 row)
```

I think this is now ready for review and merge.

rafatower:
@Algunenano you just need to look at the last commit https://github.com/CartoDB/cartodb-postgresql/pull/383/commits/d76e3ccc3e30358bd8bf2442ad0e4e917c58f248 and I can merge it into branch `federated_function`

scripts-available/CDB_FederatedServerDiagnostics.sql
```diff
@@ -81,6 +81,22 @@ $$
...
 
 
+--
+-- Get the foreign server options of a remote PG server
+--
+CREATE OR REPLACE FUNCTION @extschema@.__CDB_FS_Foreign_Server_Options_PG(server_internal name)
+RETURNS jsonb
+AS $$
+    -- See https://www.postgresql.org/docs/current/catalog-pg-foreign-server.html
+    SELECT jsonb_object(server_options) FROM (
```
Algunenano: 2 things:
- I'm not sure what's all the unnest + regex about. Could it be solved using `pg_options_to_table`?
- This might return the remote password, which we should avoid if we don't want to worry about it later (caching, logs, etc).

rafatower: > I'm not sure what's all the unnest + regex about. Could it be solved using pg_options_to_table?

this is about converting `srvoptions` from a `Foreign server specific options, as “keyword=value” strings` into a json object. I did not know about `pg_options_to_table` and I'm not sure it can be applied there, but if I can simplify it with it or some other function, I'll do.

> This might return the remote password, which we should avoid if we don't want to worry about it later (caching, logs, etc).

Ok.

rafatower: PS: the password comes in the user mapping, so that's one less concern to worry about.

Algunenano: > PS: the password comes in the user mapping, so that's one less concern to worry about.

That's true. Sorry for the noise.

scripts-available/CDB_FederatedServerDiagnostics.sql
```diff
@@ -81,6 +81,22 @@ $$
...
+        SELECT ARRAY(SELECT regexp_split_to_table(opts, '=')
+        FROM pg_foreign_server, unnest(srvoptions) opts
+        WHERE srvname = cartodb.__CDB_FS_Generate_Server_Name(input_name => 'loopback', check_existence => true)) server_options
```
rafatower: Notes to myself:

- s/'loopback'/server_internal/
- s/cartodb/@extschema@/

test/CDB_FederatedServerDiagnostics.sql
```diff
@@ -0,0 +1,75 @@
...
+\echo '%% It returns null as the postgis version if it is not installed'
+\set QUIET on
+\c cdb_fs_tester postgres
+DROP EXTENSION postgis;
+\c contrib_regression postgres
+\set QUIET off
+SELECT '1.5', cartodb.CDB_Federated_Server_Diagnostics(server => 'loopback') @> '{"postgis_version": null}'::jsonb;
+
+\echo '%% It returns the remote server options'
+SELECT '1.6', cartodb.CDB_Federated_Server_Diagnostics(server => 'loopback') @> '{"server_options": {"host": "localhost", "port": "5432", "updatable": "false", "extensions": "postgis", "fetch_size": "1000", "use_remote_estimate": "true"}}'::jsonb;
```
Algunenano: 5432 => @@PGPORT@@

PR #385, opened 2019-11-14
Federated functions permissions

Algunenano:
2 changes:
* Grant internal role explicit access over public and cartodb schemas. This is to work around some builder settings that remove permissions over those schemas to public.
* All views are now placed in an auxiliary schema:  This is to ensure that the internal role has creation permission over the destination schema.


scripts-available/CDB_FederatedServer.sql
```diff
@@ -19,6 +26,7 @@ LANGUAGE SQL IMMUTABLE PARALLEL SAFE;
 --
 -- Produce a valid DB name for servers generated for the Federated Server
 -- If check_existence is true, it'll throw if the server doesn't exists
+-- This name is also used to
```
rafatower: ??

PR #387, opened 2019-12-20
use literal substitution instead of identifier

alrocar:
Related issue https://github.com/CartoDB/support/issues/2273

I made a quick test in one of the affected users database and the fix seems to work


rafatower:
Fine by me, but most likely the problem comes from double quoting, that is, changes in the parameter passed to the function (I can tell for sure this used to work ok under different conditions).

alrocar:
This is how it's being called from cartodb (which is the origin of the issue):

```
CDB_UserDataSize('#{self.database_schema}')
```

I did some manual tests and it seems to happen when the database_schema contains `-`. I'll try to add some tests to check that to the PR.

PR #395, opened 2020-02-11
Make cartodb-postgresql python3 compatible

manmorjim:
- encoding string with python3
- adapt HTTPConnection signature to python3

### DIFF
--- a/scripts-available/CDB_Groups_API.sql
+++ b/scripts-available/CDB_Groups_API.sql
@@ -202,9 +202,9 @@ $$
 
     data_to_encode = '%s:%s' % (username, password)
     if sys.version_info[0] < 3:
-        data_encoded = base64.encodestring(data_to_encode)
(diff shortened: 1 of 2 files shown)


scripts-available/CDB_Groups_API.sql
```diff
@@ -197,19 +197,30 @@ CREATE OR REPLACE
...
     RETURNS int AS
 $$
+    python_v2 = True
     try:
-        import httplib as client
+      import httplib as client
     except:
-        from http import client
+      from http import client
+      python_v3 = False
```
Algunenano: Name should be python_v2 here

manmorjim: :man_facepalming: I changed it in cartodb/lib/sql/scripts-available instead of here. Sorry about that and good catch!

Issue #352, opened 2019-05-01
Type existence check for `_cdb_has_usable_geom_record` isn't search_path aware


nballenger:
I'm trying to get the test suite running for the SQL API repo, and one of the issues I've run into is that the function def for `_CDB_Has_Usable_Geom` in `scripts-available/CDB_CartodbfyTable.sql` fails in the test database, because the Postgres instance claims that the data type `_cdb_has_usable_geom_record` doesn't exist.

The test database is templated off the `template_postgis` database which has the cartodb-postgresql extension installed, so in fact the function and data type should be there, but because Postgres doesn't let you inherit `search_path` from the template database, they're not actually visible to the test run user.

Conceivably that wouldn't be a problem, since `scripts-available/CDB_CartodbfyTable.sql` just goes ahead and redefines the datatype if it isn't present, EXCEPT that it does so by running the following:

https://github.com/CartoDB/cartodb-postgresql/blob/74210c5b5cec34bbcba117b973d7d17ab3ad26a0/scripts-available/CDB_CartodbfyTable.sql#L637-L650

The `SELECT 1 FROM pg_type...` query on 639 succeeds because the data type does exist in `pg_catalog.pg_type`, but it isn't actually visible to the user / process because the search_path for the new database doesn't include the `cartodb` schema.

So two things:

1. If you add `AND pg_catalog.pg_type_is_visible(oid)` to that `SELECT 1` query, it will return 0 if the datatype isn't in the search_path for the database (and so will create it).
1. Seems like maybe in your Depends On section up top of CDB_CartodbfyTable.sql, you may want to include CDB_SearchPath.sql?

Algunenano:
This should have been fixed when we added the schema to all function calls and types inside the extension. In any case, I'll modify the condition to check just the extension schema and not the whole `pg_type`.

PR #400, opened 2020-11-17
CDB_RegenerateTable

Algunenano:
Adds a new function, `cartodb.CDB_RegenerateTable` that regenerates a table in place, refreshing it on disk but keeping dependencies intact.

Works with (or should):
* Data (kept as it was, although the order in disk might change)
* Triggers
* Indices
* Role permissions
* Column properties (including sequential series)

Does not work (will fail and leave everything as is):
* Views and materialized views.
* Any other direct dependencies on a table (not sure if there is any)

I've also raised the minimum PG release to 11. Prior release will probably still work, but we don't test it so they will start failing at some point.

Closes https://github.com/CartoDB/cartodb-postgresql/issues/352


scripts-available/CDB_RegenerateTable.sql
```diff
@@ -0,0 +1,82 @@
...
+RETURNS text[]
+AS $$
+    import subprocess
+    import re
+
+    query = "SELECT current_database()::text as dname"
+    rv = plpy.execute(query, 1)
+    database_name_string = str(rv[0]['dname'])
+
+    query = """SELECT concat(quote_ident(nspname), '.', quote_ident(relname)) as quoted_name
```
jgoizueta, line 15: why using a query rather than relying on fomat %I and casting the tableoid to regclass? oh, uhm, I guess that would depend on the search_path to include or not the schema... and now that I think of it, using format from plpython means running a query anyway... oh, forget it

scripts-available/CDB_RegenerateTable.sql
```diff
@@ -0,0 +1,82 @@
...
+    queries TEXT[] := @extschema@.__CDB_RegenerateTable_Get_Commands(tableoid);
+    i INTEGER;
+BEGIN
+    EXECUTE FORMAT('SELECT concat(quote_ident(nspname), ''.'', quote_ident(relname)) as quoted_name
+                        FROM pg_catalog.pg_class AS c
+                        JOIN pg_catalog.pg_namespace AS ns
+                        ON c.relnamespace = ns.oid
+                        WHERE c.oid = %L', tableoid) INTO table_name;
+
+    RAISE DEBUG '%', FORMAT('CREATE TEMPORARY TABLE %s  ON COMMIT DROP AS SELECT * FROM %s', temp_name, table_name);
```
jgoizueta, line 68: do you plan to keep these debug messages? (if so I'd generate the commands once, mostly to make sure they're kept in sync with the executed code)

scripts-available/CDB_TransformToWebmercator.sql
```diff
@@ -36,7 +36,7 @@ BEGIN
 
   -- Don't bother clipping if the geometry boundary doesn't
   -- go outside the valid extent.
-  IF latlon_input @ valid_extent THEN
+  IF @postgisschema@.geometry_within(latlon_input, valid_extent) THEN
```
jgoizueta, line 39: ?

jgoizueta (approved):
Nicely done. This may be a hack, but is a professionally engineered hack.

scripts-available/CDB_TransformToWebmercator.sql
Algunenano: CDB_TransformToWebmercator uses the postgis operator `@` which is under the schema public. Since I didn't know how to change `@` to something like `public.operator@(geometry,geometry)` I changed it to the underlying Postgis function so I could add the `@postgisschema@ in front. Without the proper schema restoring some triggers might fail since  the pg_dump options change the search_path (removing public).

scripts-available/CDB_RegenerateTable.sql
Algunenano: Uhmm, I should probably remove the logs. The dump process is even raising them to warning so it only outputs half of them :D

scripts-available/CDB_RegenerateTable.sql
Algunenano: Yeah, I wasn't sure how to do this in a more proper way but I do expect one to exist :D I just knew that doing `quote_ident(schema) + '.' + quote_ident(tablename)` should always work, independently of search_path or special characters in either of the names.

scripts-available/CDB_Overviews.sql
```diff
@@ -268,44 +268,7 @@ AS $$
...
-          WHERE ctid = ANY (
-            ARRAY[
-              (SELECT @extschema@.CDB_RandomTids(''%2$s'', %3$s))
-            ]
-          );
-      ', creation_clause, reloid, num_samples);
-    END IF;
-
-    RETURN Format('%s', overview_table_name)::regclass;
+    RAISE EXCEPTION 'Creating overviews is deprecated';
```
jgoizueta, line 271: Since you've removed the code, I guess it's, technically, "eliminated" rather than "deprecated". 😝

jgoizueta (approved):
LGTM 🏁

PR #401, opened 2020-12-02
Change `__CDB_RegenerateTable_Get_Commands` to use the caller timeout or '1min' if not set

Algunenano:
Raise the timeout to get the lock to whatever the current `statement_timeout` for the user is. To avoid blocking indefinitely, if the timeout is 0, use 1 minute instead.


manmorjim (approved):
Great! so fast :ok_hand: :rocket:

